PR #25, opened 2024-07-29
Proving density of $E_8$

grhkm21:
In this PR, we prove the {finite,periodic,lattice,} density of $E_8$.

A priori the definition of density of $X$ is $\Delta_X = \lim_{R \to \infty} \frac{\mathrm{Vol}(\mathcal{P} \cap \mathcal{B}(R))}{\mathrm{Vol}(\mathcal{B}(R)}$, where $\mathcal{P} = \bigcup_{x \in X} \mathcal{B}\left(\frac{r}{2}\right)$ and $r$ is the "separation radius" of $X$.

However, the definition is annoying to work with in our case, since $E_8$ doesn't have a nice "growing outwards" structure, whatever that means. Moreover, there is clearly a better way to compute the density of a periodic packing. If $X$ is $\Lambda$-invariant, then $\Delta_X = \frac{|X / \Lambda|\mathrm{Vol}\left(\mathcal{B}\left(\frac{r}{2}\right)\right)}{\mathrm{Covol}(\Lambda)}$.

So this PR just proves they are equal. The core idea is to look at growing circles. For more details, look at the blueprint, it exists for a reason.

Depends on: #28


SpherePacking/Basic/SpherePacking.lean
```diff
@@ -47,18 +89,49 @@ noncomputable instance PeriodicSpherePacking.instAddAction (S : PeriodicSpherePa
...
 
 noncomputable def SpherePacking.finiteDensity (S : SpherePacking d) (R : ℝ) : ℝ≥0∞ :=
   volume (S.balls ∩ ball 0 R) / (volume (ball (0 : EuclideanSpace ℝ (Fin d)) R))
 
 noncomputable def SpherePacking.density (S : SpherePacking d) : ℝ≥0∞ :=
   limsup S.finiteDensity atTop
 
+-- Here is one way to choose a basis, but I think for our purpose we can just supply one.
+-- /-- Returns a ℝ-basis of ℝ^d such that the its ℤ-span is `S.lattice`. -/
+-- noncomputable def PeriodicSpherePacking.lattice_basis (S : PeriodicSpherePacking d) :
```
thefundamentaltheor3m, line 112: Maybe we don't need to comment this out... if someone seeking to use basis-dependent lemmas has a basis in mind, they can use that, but if they just need a basis for the sake of needing one, they can use this one. What do you think?
```suggestion
noncomputable def PeriodicSpherePacking.lattice_basis (S : PeriodicSpherePacking d) :
     Basis (Module.Free.ChooseBasisIndex ℤ S.lattice) ℝ (EuclideanSpace ℝ (Fin d)) :=
   ((Zlattice.module_free ℝ S.lattice).chooseBasis).ofZlatticeBasis _ _
```

SpherePacking/Basic/SpherePacking.lean
```diff
@@ -196,3 +325,260 @@ theorem constant_eq_constant_normalized (hd : 0 < d) :
...
+  apply measure_iUnion ?_ (fun _ ↦ measurableSet_ball)
+  intro ⟨x, hx⟩ ⟨y, hy⟩ h
+  apply ball_disjoint_ball
+  simp_rw [ne_eq, Subtype.mk.injEq] at h ⊢
+  linarith [S.centers_dist' x y hx.left hy.left h]
+
+-- https://github.com/leanprover-community/mathlib4/pull/15214/files
+-- Putting it as axioms so that #print axioms will show that this should be removed
+-- TODO: remove when merged
+axiom ENNReal.tsum_const_eq' {α : Type*} (s : Set α) (c : ENNReal) :
```
thefundamentaltheor3m: Looks like it's been merged. Safe to remove?

SpherePacking/Basic/SpherePacking.lean
```diff
@@ -196,3 +325,260 @@ theorem constant_eq_constant_normalized (hd : 0 < d) :
...
+    rwa [add_sub_cancel_right] at this
+  · exact (volume_ball_pos _ (by linarith [S.separation_pos])).ne.symm
+  · exact (volume_ball_lt_top _).ne
+
+end BasicResults
+
+
+
+
+section ScratchPad
```
thefundamentaltheor3m: Maybe we can rename/reorganise this section a bit. Code looks good, just a suggestion for readability

SpherePacking/Basic/PeriodicPacking.lean
```diff
@@ -0,0 +1,746 @@
...
+section theorem_2_2
+open scoped Pointwise
+variable {d : ℕ} (S : PeriodicSpherePacking d)
+  {ι : Type*} [Fintype ι] [DecidableEq ι]
+  (D : Set (EuclideanSpace ℝ (Fin d)))
+  -- (hD_fd : IsAddFundamentalDomain S.lattice D)
+
+  -- this strengthens hD_fd.ae_covers
+  -- It's not actually necessary but it'll make the proof much much more annoying
+  (hD_unique_covers : ∀ x, ∃! g : S.lattice, g +ᵥ x ∈ D)
```
thefundamentaltheor3m: Perhaps it would be more direct to have `hD_fd` instead of `hD_unique_covers`... surely `hD_unique_covers` should follow from it...

thefundamentaltheor3m: Would `hD_fd` also imply `hD_measurable`?

SpherePacking/Basic/PeriodicPacking.lean
```diff
@@ -0,0 +1,746 @@
...
+  use ⟨v, hv⟩, hv'.left, ?_
+  intro ⟨y, hy⟩ hy'
+  have := hv'.right ⟨y, ?_⟩ hy'
+  rwa [Subtype.ext_iff] at this ⊢
+  rw [S.basis_Z_span]
+  exact hy
+
+section finiteDensity_limit
+
+/- TODO: consider moving this section. -/
```
thefundamentaltheor3m, line 645: I think we can leave it where it is, because after all it does concern periodic packings. Perhaps we can just rename it to make it extra clear that this is where the main result, `PeriodicSpherePacking.density_eq`, lives

SpherePacking/Basic/SpherePacking.lean
grhkm21: I don't think it's necessary, it's rare you would need a $R$ basis induced by an *arbitrary* lattice $Z$-basis, and in that case you can just take one as an argument (like the rest of the file)

SpherePacking/Basic/SpherePacking.lean
grhkm21: Yes, I will bump Mathlib and remove it

SpherePacking/Basic/PeriodicPacking.lean
grhkm21: Well the reason I added this argument is precisely because it does *not* follow from `hD_fd` - that's what I mean for "this strengthens ...". If you read the definition it should be obvious why.

PR #45, opened 2024-11-08
bunch of sorrys for modular forms.

CBirkbeck:


SpherePacking/ModularForms/Eisenstein.lean
```diff
@@ -5,27 +5,160 @@ The purpose of this file is to define the Eisenstein series we are interested in
...
+def E (k : ℤ) (hk : 3 ≤ k) : ModularForm (CongruenceSubgroup.Gamma ↑1) k :=
+  (1/2) • eisensteinSeries_MF hk standardcongruencecondition /-they need  1/2 for the
+    normalization to match up (since the sum here is taken over coprime integers).-/
 def E₆ : ModularForm (CongruenceSubgroup.Gamma ↑1) 6 :=
-  eisensteinSeries_MF (by norm_num) standardcongruencecondition
+  (1/2) • eisensteinSeries_MF (by norm_num) standardcongruencecondition
+
+def S0 : Set ℤ := {0}ᶜ
+
+def G₂' : ℍ → ℂ := fun z => ∑' (m : ℤ), (∑' (n : ℤ), 1 / (m * z + n) ^ 2) --hmm is this right?
```
seewoo5, line 38: Does this exclude $(m, n) = (0, 0)$? Also, in that case, I guess it needs to be multiplied by $(1/2)$ to get 1 for the constant term of the Fourier expansion (or if you intended this to be $2E_2$ then I think it is correct)

CBirkbeck: No it doesnt exclude it since in that case its just 1/0 which is just 0. And yes if you look on line 135 where E_2 is defined I took the 1/2 out. But more seriously I think this definition is probably wrong, since it would need us to know the outer sum is absolutely convergent, which I cant see being true. I can prove the inner sum is, but not the outer sum (which would still be weaker than asking the double sum to be abs conv. which is false). This is why I have also defined G_2 and I think thats closer to the definition we want.

CBirkbeck: The reason I made it was because at somepoint there was some discussion on discord about both sums being abs convergent, but now that I look at it I can't see why.

PR #63, opened 2025-05-28
define normalized derivative and Serre derivative

seewoo5:
Define $$D = (2 \pi i)^{-1} \frac{\mathrm{d}}{\mathrm{d}z}$$ and $$\partial_k = D - \frac{k}{12} E_2$$. Preparation for proving Ramanujan's formula.


SpherePacking/ModularForms/Derivative.lean
```diff
@@ -0,0 +1,105 @@
...
+
+
+noncomputable def serre_D (k : ℂ) (F : ℍ → ℂ) : ℍ → ℂ := λ (z : ℍ) => D F z - k * 12⁻¹ * E₂ z * F z
+
+/--
+Basic properties of Serre derivative: linearity, Leibniz rule, etc.
+-/
+theorem serre_D_add (k : ℤ) (F G : ℍ → ℂ) (z : ℍ) :
+    serre_D k (F + G) z = serre_D k F z + serre_D k G z
+  := by
```
b-mehta:
```suggestion
    serre_D k (F + G) z = serre_D k F z + serre_D k G z := by
```
and similarly below

SpherePacking/ModularForms/Derivative.lean
```diff
@@ -0,0 +1,101 @@
...
+/--
+Basic properties of derivatives: linearity, Leibniz rule, etc.
+-/
+theorem D_add (F G : ℍ → ℂ) (z : ℍ) : D (F + G) z = D F z + D G z := by sorry
+
+theorem D_smul (c : ℂ) (F : ℍ → ℂ) (z : ℍ) : D (c • F) z = c * D F z := by sorry
+
+theorem D_mul (F G : ℍ → ℂ) (z : ℍ) : D (F * G) z = F z * D G z + G z * D F z := by sorry
+
+theorem D_const (c : ℂ) (z : ℍ) : D (Function.const _ c) z = 0 := by sorry
```
CBirkbeck: So I think you are going to need these to be for F,G that are differentiable on the upper half plane, as otherwise they might be false, as, for example, deriv (f + g) = deriv f + deriv g is only true for differentiable functions. Its probably a good idea to prove at least one of these to make sure you have the right hypothesis going forward.

seewoo5: I added proofs except for `D_mul`.

seewoo5:
@b-mehta @CBirkbeck Could you review this one? There are also some updates in blueprint.

SpherePacking/ModularForms/Derivative.lean
```diff
@@ -0,0 +1,274 @@
...
+open Real Complex CongruenceSubgroup SlashAction SlashInvariantForm ContinuousMap
+
+open scoped ModularForm MatrixGroups Manifold
+
+/-!
+Definition of (Serre) derivative of modular forms.
+Prove Ramanujan's formulas on derivatives of Eisenstein series.
+-/
+
+noncomputable def E₄_fun : ℍ → ℂ := E₄.toFun
```
CBirkbeck: Do you really need these? why not just use the .toFun?

SpherePacking/ModularForms/Derivative.lean
```diff
@@ -0,0 +1,274 @@
...
+    D (Function.const _ c) z
+    _ = (2 * π * I)⁻¹ * deriv (Function.const _ c ∘ ofComplex) z := by rfl
+    _ = (2 * π * I)⁻¹ * 0 := by rw [h]
+    _ = 0 := by ring_nf
+
+
+/--
+Serre derivative of weight `k`.
+-/
+noncomputable def serre_D (k : ℂ) : (ℍ → ℂ) → (ℍ → ℂ) := λ (F : ℍ → ℂ) => (λ z => D F z - k * 12⁻¹ * E₂ z * F z)
```
CBirkbeck, line 138: Im guessing eventually you'll want to make this into a map between spaces of modular forms?

b-mehta (commented):
LGTM, modulo Chris' suggestion.

SpherePacking/ModularForms/Derivative.lean
seewoo5: I thought this makes writing codes simpler, but I found that it doesn't ;D I replaced with `E₄.toFun` back.

SpherePacking/ModularForms/Derivative.lean
seewoo5: Yes exactly, this has to be upgraded in that sense later.

PR #115, opened 2025-06-22
chore: update build workflow actions and settings

pitmonticone:
Disables GitHub cache in the Lean build step and updates the `docgen-action` to a newer commit. These changes ensure the workflow uses the latest tools and avoids potential cache-related issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,9 +34,11 @@ jobs:
 
       - name: Build the project
         uses: leanprover/lean-action@f807b338d95de7813c5c50d018f1c23c9b93b4ec # v1.2.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -34,9 +34,11 @@ jobs:
 
       - name: Build the project
         uses: leanprover/lean-action@f807b338d95de7813c5c50d018f1c23c9b93b4ec # v1.2.0
+        with:
+          use-github-cache: false
```
pitmonticone, line 38: This is necessary to avoid caching 1.9 GB of build data on every CI run.

b-mehta (approved):

PR #119, opened 2025-06-27
redo the E8 lattice file

b-mehta:
The old file had a number of sorries, some of which were false and others of which missed out large proofs. This PR (a work-in-progress) redoes the file to fix both of these problems.


SpherePacking/Basic/E8.lean
b-mehta, line 1: I'm changing the copyright on this file since it was essentially redone from scratch, with only small parts taken from the old file.

b-mehta: As such, it's better to review this file as a new file rather than a changed file.

b-mehta:
> I think the blueprint refs need to be updated...

This should be done now!

PR #126, opened 2025-06-30
fill in periodic packing sorries

b-mehta:


SpherePacking/ForMathlib/ENat.lean
```diff
@@ -24,11 +24,11 @@ theorem ENat.tsum_const {α : Type*} (c : ENat) :
       _ = t.card * 1 := by simp
       _ ≤ t.card * c := mul_le_mul_left' (ENat.one_le_iff_ne_zero.2 hc) _
 
+theorem ENat.tsum_set_const {α : Type*} (s : Set α) (c : ENat) :
```
seewoo5, line 27: Just a minor question - is there a standard for the order of the theorems (in mathlib)?

SpherePacking/ForMathlib/Encard.lean
```diff
@@ -1,9 +1,233 @@
...
+
+import Mathlib.Data.ENat.Lattice
+import Mathlib.Data.Set.Card
+import Mathlib.Topology.Instances.Discrete
+import Mathlib.Topology.Order.T5
+import Mathlib.Order.Interval.Set.WithBotTop
+import Mathlib.Order.Filter.Pointwise
+import Mathlib.Topology.Algebra.Monoid.Defs
+import Mathlib.Topology.Algebra.InfiniteSum.Constructions
+import Mathlib.Topology.Algebra.InfiniteSum.Order
```
seewoo5:
```suggestion
import Mathlib.Data.ENat.Lattice
import Mathlib.Topology.Algebra.InfiniteSum.Order
import Mathlib.Topology.Order.T5
import SpherePacking.ForMathlib.ENat
```
`#min_imports` says those are enough (and I also checked myself)!

SpherePacking/ForMathlib/ENat.lean
b-mehta: In this case I think the order tsum_const / tsum_set_const // tsum_one / tsum_set_one makes sense, whereas the other order felt a little arbitrary.  So essentially that's just grouping similar lemmas where possible

seewoo5 (approved):
Thanks!

PR #136, opened 2025-07-16
[WIP] Add linter

seewoo5:


SpherePacking/ModularForms/eta.lean
```diff
@@ -17,7 +17,6 @@ noncomputable def η (z : ℂ) := cexp (2 * π * Complex.I * z / 24) * ∏' (n :
     (1 - cexp (2 * π * Complex.I * (n + 1) * z))
 
 
-/-this is being PRd-/
```
b-mehta, line 20: Can we keep this comment around in some form?

SpherePacking/ModularForms/cotangent.lean
```diff
@@ -193,7 +193,7 @@ theorem logDeriv_sin_div (z : ℂ) (hz : z ∈ ℂ_ℤ) :
   · simp only [ne_eq, mul_eq_zero, ofReal_eq_zero, not_or]
     refine ⟨Real.pi_ne_zero, integerComplement.ne_zero hz⟩
 
-/--The term in the infinite series expansion of cot. -/
+/- The term in the infinite series expansion of cot. -/
```
b-mehta:
```suggestion
/-- The term in the infinite series expansion of cot. -/
```

b-mehta (approved):

b-mehta:
This looks nicer than I expected, thanks!

SpherePacking/ModularForms/eta.lean
seewoo5: Probably @CBirkbeck knows? I removed it because it is actually PRd (at least no sorry in the file).

SpherePacking/ModularForms/cotangent.lean
seewoo5: All the one-line dostring starts with `/--`? I may probably update these, too

b-mehta: Docstrings should start with `/--`, yeah. If they start with `/-` then they're just comments.

SpherePacking/ModularForms/eta.lean
b-mehta: If it's actually PR'd, then we should remove the lemma altogether

b-mehta:
Oops, I didn't realise there's lots of warnings still left to fix!

SpherePacking/ModularForms/eta.lean
CBirkbeck: yeah it has been PRd, but we might need to update the file with the new version. I can sort it out though afterwards if thats easier

seewoo5:
> Thanks for working on this; I'm always happy about code clean-up (and particularly in a project like this that I'm excited about).
> 
> If I were reviewing this (which I don't, I'm an outsider), I would suggest splitting this into several parts. The whitespace fixes can land independently, for instance --- so looking at the file removals can happen independently.

I agree - I can actually do if that makes reviewing better for other (for the same reason I haven't fixed other warnings, but as you said further PR split is possible). If it is okay to not to separate PR, at least I'd emphasize that the actual *errors* are from non-spaces of one-line docstrings and unused files, while I also fixed some *warnings* for other spaces.

PR #137, opened 2025-07-20
Fix lint warnings for 100 chars limit

seewoo5:


SpherePacking/ForMathlib/InvPowSummability.lean
thefundamentaltheor3m, line 1: Line 238 also seems to be > 100 chars long, and unnecessary from the looks of it, so I say we get rid of it...

thefundamentaltheor3m (commented):
Looks good! I hope I didn't miss anything - line 238 in `SpherePacking.ForMathlib.InvPowSummability` seems to me to be the only remaining one with >100 chars... could you recheck @seewoo5? Thank you!

SpherePacking/ForMathlib/InvPowSummability.lean
seewoo5: Removed!

PR #139, opened 2025-07-28
bump

CBirkbeck:


SpherePacking/ModularForms/iteratedderivs.lean
```diff
@@ -17,7 +17,7 @@ theorem upper_ne_int (x : ℍ) (d : ℤ) : (x : ℂ) + d ≠ 0 :=
   by
   by_contra h
   rw [add_eq_zero_iff_eq_neg] at h
-  have h1 : 0 < (x : ℂ).im := by simp only [coe_im]; exact im_pos x
+  have h1 : 0 < (x : ℂ).im := by simp; exact im_pos x
```
thefundamentaltheor3m, line 20: I think we should keep the `simp only` version to avoid having a non-terminal `simp`. (On a side note, we have a few non-terminal `simp`s scattered around the repo that we need to get rid of at some point.)

SpherePacking/ModularForms/iteratedderivs.lean
```diff
@@ -27,35 +27,47 @@ theorem aut_iter_deriv (d : ℤ) (k : ℕ) :
       (fun t : ℂ => (-1) ^ k * k ! * (1 / (t + d) ^ (k + 1))) {z : ℂ | 0 < z.im} := by
   intro x hx
   induction' k with k IH generalizing x
-  · simp
+  simp only [iteratedDerivWithin_zero, pow_zero, Nat.factorial_zero, one_mul]
+  simp  at *
```
thefundamentaltheor3m, line 31: Again, we need to get rid of this non-terminal `simp`

SpherePacking/ModularForms/logDeriv_lems.lean
```diff
@@ -1,22 +1,61 @@
...
+theorem logDeriv_tprod_eq_tsum2 {s : Set ℂ} (hs : IsOpen s) (x : s) (f : ℕ → ℂ → ℂ)
+    (hf : ∀ i, f i x ≠ 0)
+    (hd : ∀ i : ℕ, DifferentiableOn ℂ (f i) s) (hm : Summable fun i ↦ logDeriv (f i) ↑x)
+    (htend : MultipliableLocallyUniformlyOn f s) (hnez : ∏' (i : ℕ), f i ↑x ≠ 0) :
+    logDeriv (∏' i : ℕ, f i ·) x = ∑' i : ℕ, logDeriv (f i) x := by
+    have h2 := Summable.hasSum hm
+    rw [Summable.hasSum_iff_tendsto_nat hm] at h2
+    apply symm
+    rw [← Summable.hasSum_iff hm]
+    rw [Summable.hasSum_iff_tendsto_nat hm]
```
thefundamentaltheor3m, line 24: Could potentially merge lines 23 and 24?

SpherePacking/ModularForms/logDeriv_lems.lean
```diff
@@ -102,11 +143,11 @@ lemma logDeriv_q_expo_summable (r : ℂ) (hr : ‖r‖ < 1) : Summable fun n : 
     apply le_trans this.le
     norm_cast
 
-lemma func_div (a b c d : ℂ → ℂ) (x : ℂ) (hb : b x ≠ 0) (hd : d x ≠ 0) :
+lemma func_div (a b c d : ℂ → ℂ) (x : ℂ) (hb : b x ≠ 0) (hd :  d x ≠ 0) :
      (a / b) x = (c /d) x ↔ (a * d) x = (b * c) x := by
   constructor
   intro h
-  simp only [Pi.div_apply] at h
+  simp at *
```
thefundamentaltheor3m, line 150: Again we should get rid of this non-terminal `simp`

SpherePacking/ModularForms/logDeriv_lems.lean
```diff
@@ -206,6 +245,8 @@ lemma logDeriv_eqOn_iff (f g : ℂ → ℂ) (s : Set ℂ) (hf : DifferentiableOn
     simp_rw [logDeriv_apply]
     have HJ := deriv_EqOn_congr s hz hs2 hx
     rw [HJ, h]
-    nth_rw 1 [show z • g = fun x => z • g x by rfl]
+
+    rw [deriv_const_smul]
     simp
```
thefundamentaltheor3m, line 250: Need to get rid of non-terminal `simp` here

thefundamentaltheor3m (changes requested):
Looks great, thanks Chris! I did notice a few non-terminal `simp`s, and I think there's a few more in our repo. Do you think we should fix them here or make a separate PR to do all of them in one go?

CBirkbeck:
> Looks great, thanks Chris! I did notice a few non-terminal `simp`s, and I think there's a few more in our repo. Do you think we should fix them here or make a separate PR to do all of them in one go?

So all of this will be cleaned up before going to mathlib (which is something I am working on now), so I don't think we need to worry about it too much since most of it will change quite a lot in this process. But we can do this if you want, but maybe in another PR since I expect there are lots of these in the repo!

PR #143, opened 2025-08-09
chore: lint (partly)

YaelDillies:


SpherePacking/ModularForms/IsCuspForm.lean
```diff
@@ -4,27 +4,27 @@ import SpherePacking.ModularForms.qExpansion_lems
 open ModularForm UpperHalfPlane TopologicalSpace Set MeasureTheory intervalIntegral
   Metric Filter Function Complex MatrixGroups
 
-open scoped Interval Real NNReal ENNReal Topology BigOperators Nat Classical
+open scoped Interval Real NNReal ENNReal Topology BigOperators Nat
```
seewoo5, line 7: Just out of curiosity, is there a thing like removing unnecessary open similar to `#min_imports`?

seewoo5 (approved):

SpherePacking/ModularForms/IsCuspForm.lean
YaelDillies: I believe Rothgang is working on a linter that would flag those unused `open`s. You could ask on Zulip.

PR #149, opened 2025-09-09
chore: fix linter warnings in ModularForms/Eisenstein

dwrensha:
Currently when I run `lake build` on this project, it reports many linter warnings.

This PR addresses the warnings that originate in the file `ModularForms/Eisenstein.lean`.

PR #150, opened 2025-09-12
fix linter warnings in four more files

dwrensha:
Fixes linter warnings in:
* `MagicFunction/a/Basic.lean`
* `ModularForms/Derivative.lean`
* `ModularForms/DimensionFormulas.lean`
* `ModularForms/Eisensteinqexpansions.lean`

As with #149, none of these changes conflict with #143.


b-mehta (commented):
Thanks!

PR #166, opened 2025-10-22
fix step_11 in PolyFourierCoeffBound

dwrensha:
Eliminates two sorries that were added in 407ac19f.

What changed is the ordering of the goals emitted by `gcongr`.

### DIFF
--- a/SpherePacking/MagicFunction/PolyFourierCoeffBound.lean
+++ b/SpherePacking/MagicFunction/PolyFourierCoeffBound.lean
@@ -328,11 +328,10 @@ private lemma step_11 :
     norm_cast at this
     exact summable_real_norm_mul_geometric_of_norm_lt_one hnorm this
   · next j =>
-      have : -π * ↑j / 2 = -π * ↑j * (1 / 2) := by rw [mul_one_div]
-      rw [this]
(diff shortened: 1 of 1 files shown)


SpherePacking/MagicFunction/PolyFourierCoeffBound.lean
```diff
@@ -325,21 +323,16 @@ private lemma step_11 :
...
-  -- rw [this]
-  -- simp at *
-  -- have hz2 := hz.le
-  -- gcongr
+  · next j =>
+      have : -π * ↑j / 2 = -π * ↑j * (1 / 2) := by rw [mul_one_div]
+      rw [this]
+      simp at *
+      have hz2 := hz.le
+      gcongr
```
b-mehta:
```suggestion
    have : -π * j / 2 = -π * j * (1 / 2) := by rw [mul_one_div]
    rw [this]
    simp at *
    have hz2 := hz.le
    gcongr
```

dwrensha: Done in 0ee71b081906a7c541c1062154a14adfffd1cc8e.

PR #167, opened 2025-10-27
feat(MagicFunction/b): Some API

thefundamentaltheor3m:
In this PR we develop some API for the $-1$-eigenfunction $b$. We also make some improvements/optimisations to previously written code about the `Ψ` functions and their slash action relations.

### DIFF
--- a/SpherePacking/MagicFunction/IntegralParametrisations.lean
+++ b/SpherePacking/MagicFunction/IntegralParametrisations.lean
@@ -179,7 +179,7 @@ lemma neg_inv_one_add_eq_ST (z : ℍ) :
   norm_cast
 
 lemma neg_inv_eq_S_coe (z : ℍ) :
-    -1 / (z) = UpperHalfPlane.coe (S • z) := by
+    -1 / z = UpperHalfPlane.coe (S • z) := by
   rw [specialLinearGroup_apply]
   simp_all [S_eq]
 


SpherePacking/MagicFunction/IntegralParametrisations.lean
```diff
@@ -107,24 +108,91 @@ end UpperHalfPlane
...
+  exact ((S * T) • z).2
+
+lemma neg_inv_one_add_eq_ST (z : ℍ) :
+    ⟨-1 / ((z : ℂ) + 1), neg_inv_one_add_mem z⟩ = (S * T) • z := by
+  apply UpperHalfPlane.ext
+  rw [← neg_inv_one_add_eq_ST_coe]
+  norm_cast
+
+lemma neg_inv_eq_S_coe (z : ℍ) :
+    -1 / (z) = UpperHalfPlane.coe (S • z) := by
```
seewoo5:
```suggestion
    -1 / z = UpperHalfPlane.coe (S • z) := by
```
maybe?

SpherePacking/MagicFunction/b/Basic.lean
```diff
@@ -10,36 +10,115 @@ import SpherePacking.MagicFunction.IntegralParametrisations
...
+
+namespace MagicFunction.b.ComplexIntegrands
+
+def Ψ₁' : ℂ → ℂ := fun z ↦ (ψT' z) * cexp (π * I * r * z)
+
+def Ψ₂' : ℂ → ℂ := fun z ↦ (ψT' z) * cexp (π * I * r * z)
+
+def Ψ₃' : ℂ → ℂ := fun z ↦ (ψT' z) * cexp (π * I * r * z)
+
+def Ψ₄' : ℂ → ℂ := fun z ↦ (ψT' z) * cexp (π * I * r * z)
```
seewoo5, line 27: These four are all the same, right (but the ones below without `'` are different)? Just for check.

SpherePacking/MagicFunction/IntegralParametrisations.lean
thefundamentaltheor3m: Agreed - good catch

SpherePacking/MagicFunction/b/Basic.lean
thefundamentaltheor3m: Yes - these ones are the same. The reason for the repetition is that each is the integrand for a different integral, and if someday we decide to rewrite the integrands in a different form (because there are many equivalent forms) then it would be a lot easier if we have separate functions defined for each of them.

The ones without the `'` are these ones composed with parametrisations. So the ones with `'` take in a real input, whereas these ones take in a complex input.

PR #171, opened 2025-11-01
Update blueprint proofs (Schwartzness, positivity of L_1,0)

seewoo5:
Update proofs in blueprint:
- Schwartzness of a(r) and b(r): follow the argument in [@thefundamentaltheor3m 's thesis](https://thefundamentaltheor3m.github.io/M4R_Thesis/main.pdf)
- Positivity of $\mathcal{L}_{1,0}$: do not need to compute first $q$-expansion, only need to know vanishing order of $F$ and $G$ (the argument appears in the most recent version of my paper).

### DIFF
--- a/blueprint/src/subsections/construct-a-b.tex
+++ b/blueprint/src/subsections/construct-a-b.tex
@@ -225,7 +225,7 @@ Note that we can take the constants $C_0$, $C_{-2}$, and $C_{-4}$ as
 Combined with Lemma~\ref{lem:integral-bound}, this shows that the integrals $I_1$, $I_3$, and $I_5$ decay faster than any polynomial as $r \to \infty$.
 For the integrals $I_2$, $I_4$, and $I_6$, it is easier to see this since the contours are not touching the real axis.
 \begin{lemma}\label{lem:bound-I2-I4-I6}
-    There exists $C_1, C_2 > 0$ such that for all $r \geq 0$,
+    There exist $C_1, C_2 > 0$ such that for all $r \geq 0$,
     \begin{equation}
         |I_2(r)|, |I_4(r)| \leq C_1 e^{-\pi r}
     \end{equation}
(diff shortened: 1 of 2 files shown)


blueprint/src/subsections/construct-a-b.tex
```diff
@@ -146,19 +146,25 @@
...
+Instead of appealing the fact, we follow more elementary argument by Slipper \cite[Proposition 4.4.3]{Slipper}.
+The integral can be break up into two parts as
+\[
+\int_0^\infty e^{-2 \pi t} e^{\pi r^2 / t} \dd t = \int_0^{r / \sqrt{2}} e^{-\pi (2t + r^2 / t)} \dd t + \int_{r / \sqrt{2}}^{\infty} e^{-\pi (2t + r^2 / t)} \dd t
+\]
+and each part can be bounded as
+\begin{align}
+    \int_0^{r /\sqrt{2}} e^{-\pi(2t + r^2 / t)} \dd t &\le \int_0^{r / \sqrt{2}} e^{-\pi r^2 / t} \dd t = \int_{\sqrt{2} / r}^{\infty} e^{-\pi r^2 t} \frac{\dd t}{t^2} \dd t \le \frac{r^2}{2} \int_{\sqrt{2} / r}^{\infty} e^{-\pi r^2 t} \dd t = \frac{e^{-\sqrt{2} \pi r}}{2 \pi}, \\
+    \int_{r / \sqrt{2}}^{\infty} e^{-\pi (2t + r^2 / t)} \dd t &\le \int_{r / \sqrt{2}}^{\infty} e^{-2 \pi t} \dd t = \frac{e^{-\sqrt{2} \pi r}}{2 \pi},
+\end{align}
```
b-mehta: @thefundamentaltheor3m I think this is exactly the argument I sketched on the whiteboard a few months back?

blueprint/src/subsections/construct-a-b.tex
```diff
@@ -326,97 +432,184 @@
...
-    \psi_S &= -128 \frac{H_3 + H_2}{H_4^2} - 128 \frac{H_2 - H_4}{H_3^2} \\
-    &= -128 \frac{H_3^2 (H_2 - H_4) + H_4^2 (H_2 - H_4)}{H_3^2 H_4^2} \\
-    &= -128 \frac{(H_2 + H_4)^2 (2H_2 + H_4) + H_4^2(H_2 + H_4)}{H_3^2 H_4^2} \\
-    &= -128 \frac{H_2 (2H_2^2 + 5 H_2 H_4 + 5 H_4^2)}{H_3^2 H_4^2} \\
-    &= -128 \frac{H_2^3 (2H_2^2 + 5 H_2 H_4 + 5 H_4^2)}{H_2^2 H_3^2 H_4^2} \\
-    &= - \frac{1}{2} \frac{H_2^3 (2H_2^2 + 5 H_2 H_4 + 5 H_4^2)}{\Delta}.
+    |J_1(r)|, |J_3(r)|, |J_5(r)| &\le C \int_1^{\infty} e^{-\pi s} e^{\pi r / s}\, \dd s.
 \end{align}
-\end{proof}
+\end{lemma}
```
seewoo5, line 551: I simply omitted proofs of this lemma; I'm not sure how much details should I add here.

blueprint/src/subsections/construct-a-b.tex
```diff
@@ -326,97 +432,184 @@
...
+% \begin{lemma}\label{lemma: psi I psi T psi S explicit}\uses{def:psiI-psiT-psiS}\lean{ψI_eq}
+% More explicitly, we have
+% \begin{align}
+%     \psi_I(z) &= 128 \frac{H_3(z) + H_4(z)}{H_2(z)^2} + 128 \frac{H_4(z) - H_2(z)}{H_3(z)^2} \label{eqn:psiI-explicit} \\
+%     \psi_T(z) &= 128 \frac{H_3(z) + H_4(z)}{H_2(z)^2} + 128 \frac{H_2(z) + H_3(z)}{H_4(z)^2} \label{eqn:psiT-explicit} \\
+%     \psi_S(z) &= -128 \frac{H_2(z) + H_3(z)}{H_4(z)^2} - 128 \frac{H_2(z) - H_4(z)}{H_3(z)^2} \label{eqn:psiS-explicit}
+% \end{align}
+% \end{lemma}
+
+% \begin{lemma}\label{lemma:psiI-psiT-psiS-fourier}\uses{lemma: psi I psi T psi S explicit}
```
seewoo5, line 474: I think we don't need this many q-expansions. At least, only the order of the first coefficient is enough to get bounds of \psi's.

blueprint/src/subsections/construct-a-b.tex
thefundamentaltheor3m: Looks like it, but we came up with an even simpler way of doing it where we combine the fact that $e^{-\pi r^2 / r}$ decays rapidly in $r^2 / t$ with some facts about the Gamma function (that already exist in Mathlib) to show that this integral decays faster than any inverse power of r. See [Lemma 4.2.5](https://thefundamentaltheor3m.github.io/M4R_Thesis/main.pdf)

PR #157, opened 2025-09-29
Poisson Summation

CoolRmal:

PR #173, opened 2025-11-02
Add proofs by gauss

gauss-math-inc:
This PR contributes 5 proofs for previously incomplete theorems automatically generated and optimized by Gauss, Math Inc's autoformalization agent.

Theorems solved:
- `coordinateEmbedding₁₂_smooth`
- `fderiv_coordinateEmbedding₁₂_hasTemperateGrowth`
- `step_10`
- `perm_I₆`
- `I₆'_bounding_eq`

### DIFF
--- a/SpherePacking/MagicFunction/a/Eigenfunction.lean
+++ b/SpherePacking/MagicFunction/a/Eigenfunction.lean
@@ -14,6 +14,23 @@ namespace MagicFunction.a.Fourier
 
 section Integral_Permutations
 
+lemma fourier_involution {V : Type*} [NormedAddCommGroup V] [InnerProductSpace ℝ V]
+    [FiniteDimensional ℝ V] [MeasurableSpace V] [BorelSpace V] {E : Type*} [NormedAddCommGroup E]
+    [NormedSpace ℂ E] [CompleteSpace E] (f : 𝓢(V, E)) :
+    fourierTransformCLE ℂ (fourierTransformCLE ℂ f) = fun x => f (-x) :=
+by
+  ext x; change 𝓕 (𝓕 f) x = f (-x)
+  simpa [Real.fourierIntegralInv_eq_fourierIntegral_neg, neg_neg] using
+    congrArg (fun g : V → E => g (-x))
+      (f.continuous.fourier_inversion f.integrable ((fourierTransformCLE ℂ) f).integrable)
+
+lemma radial_inversion {V : Type*} [NormedAddCommGroup V] [InnerProductSpace ℝ V]
+    [FiniteDimensional ℝ V] [MeasurableSpace V] [BorelSpace V] {E : Type*} [NormedAddCommGroup E]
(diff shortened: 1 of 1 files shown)


SpherePacking/MagicFunction/a/Eigenfunction.lean
thefundamentaltheor3m, line 1: The way I think we should really be proving `perm_I₆` is by applying to `perm_I₅` the more general fact that if $f$ is radial and Schwartz, the Fourier transform applied twice to $f$ just gives $f$..

The good news is that the proof Gauss has come up with for `perm_I₆` seems to contain a proof that this is true in the case of $I_5$. It shouldn't be hard to generalise this to an arbitrary $f$. If we can do that and then apply that to prove `perm_I₆` in one line that would be even better.

SpherePacking/ForMathlib/PoissonSummation/SchwartzMap.lean
thefundamentaltheor3m, line 1: You should know that the first two results are now defunct: they were part of an older attempt to prove Poisson Summation and we're now doing it differently (see #157). Once that PR gets to a better state, we will probably remove our existing Poisson Summation files to reduce clutter.

thefundamentaltheor3m (changes requested):
These are great, thank you! If we can generalise the `perm_I₆` proof as outlined in my comment, we can go ahead and merge these changes. Thank you very much Jesse and team, this is really impressive. We're very excited to have Gauss involved in this project!

SpherePacking/MagicFunction/a/Eigenfunction.lean
dwrensha: Gauss's proof of `perm_I₆` does depend to some degree on unfolding the definition of `I₆`.

To better understand it, I tried expanding it a bit:

```lean
  calc
    fourierTransformCLE ℂ I₆
        = (SchwartzMap.fourierTransformCLE ℂ).symm I₆ := by
          ext x
          have h := congrArg (fun f => f x)
            (Real.fourierIntegralInv_eq_fourierIntegral_comp_neg
              (I₆ : EuclideanSpace ℝ (Fin 8) → ℂ))
          dsimp only [I₆] at h
          simp only [schwartzMap_multidimensional_of_schwartzMap_real, compCLM_apply,
            Function.comp_apply, norm_neg] at h
          simp [SchwartzMap.fourierTransformCLE_symm_apply,
                SchwartzMap.fourierTransformCLE_apply]
          exact h.symm

```

The `schwartzMap_multidimensional_of_schwartzMap_real`, `compCLM_apply`, and `norm_neg` lemmas all pertain to dealing with the particular form that `I₆` takes.

This is all to say: extracting a general proof from this might not be as trivial as you were hoping.

dwrensha: Ah, I suppose that the lines 
```
          dsimp only [I₆] at h
          simp only [schwartzMap_multidimensional_of_schwartzMap_real, compCLM_apply,
            Function.comp_apply, norm_neg] at h
```

correspond to proving that the function is radial, i.e. depends only on the norm of the argument.

PR #180, opened 2025-11-10
perf(CohnElkies): LPBound

ster-oc:
This PR is a refactor of `LPBound` with complex partial order. This reduces the verbosity of statements and the number of hypothesis previously required. All proofs are slight adaptations of the old ones with the advantage of being able to avoid the "coercion gymnastics" `ℝ≥0∞ → ℝ≥0 → ℝ → ℂ`

The previous version of the document is appended untouched below, as a comment.

Depends on : #224


ster99:
Cohn-Elkies has been refactored! 
All the lemmas/theorems/definitions/structures used in `LPBound` have all been put (if already present, I copied them) in `Prereqs` under the section `prerequisites`. 

Now, I am opening a different PR where I will be dealing with proving and sorting everything that `Prereqs` contains. 
In this way, we can delete `Prereqs` as per its docstring.

PR #193, opened 2025-11-15
Statements of theorems for (quasi)modular forms $F$ and $G$

seewoo5:
Closes #185. This PR is mostly for the statements, but there are some proofs, too. There can be more refined steps, which will be added subsequently while proving the inequality. Also, changed the order in `D_mul` for consistency.

~#196 needs to be merged first.~

PR #196, opened 2025-11-15
Update definition of `ResToImagAxis.Real/Pos/EventuallyPos`

seewoo5:
While I was working on PR #193, I realized the new definitions here seems better to work with.


CBirkbeck (approved):
LGTM

PR #222, opened 2025-12-02
feat(MagicFunction/PolyFourierCoeffBound): prove `norm_φ₀_le`

LorenzoLuccioli:
This PR proves the lemma `norm_φ₀_le`, which establishes the exponential decay bound for the quasimodular form `φ₀`.

The proof required developing several auxiliary lemmas to set up the Fourier coefficient machinery, including bounds showing the coefficients have polynomial growth, summability results, and the connection between the squared Eisenstein expression and its Fourier series. These ingredients then allowed applying the main bound theorem.

Co-authored-by: Aristotle Harmonic [aristotle-harmonic@harmonic.fun](mailto:aristotle-harmonic@harmonic.fun).

Original output by @Aristotle-Harmonic: https://gist.github.com/pitmonticone/4bc1952449535fe666350a0b4feaed17.


b-mehta:
Iirc we decided that we should delete the lemma E₂_mul_E₄_sub_E6, and instead ask for a proof (ai or otherwise) using the lemmas in the mentioned file. Essentially the proof here is 'cheating' since it uses a strong result in a suboptimal way, except we hadn't made this clear in the repo or given aristotle the chance to do the right thing.

PR #227, opened 2025-12-08
Initial version with proof from Aristotle of Whitney lemma on smooth e…

mattcushman:
…ven functions

This includes a proof due to Aristotle/Harmonic that a smooth even function can be written as a smooth function composed with $x \mapsto x^2$.  This resolves a sorry in `Function.Even.eq_smooth_comp_sq_of_smooth`.  A small amount of preliminary golfing has been done, but the code would benefit from much more.


thefundamentaltheor3m:
I turned this into a regular PR because I wanted to run CI on it. There are a number of goals that have been "closed" by `exact?`, which does not actually close goals!

PR #237, opened 2025-12-12
Outline for Integrability

thefundamentaltheor3m:
In this PR, we create a framework for proving that the integrands making up $$a$$ are integrable.

1. In `SpherePacking.MagicFunction.a.Basic`, we define functions as `Φ₁'...Φ₆'`. We also define `Φ₁...Φ₆` as `Φ₁'...Φ₆'` composed with the appropriate parametrisations. Finally, we define the integrals `I₁...I₆` in terms of `Φ₁...Φ₆`.
2. In `SpherePacking.MagicFunction.a.Integrability.ComplexIntegrands`, for all $$j = 1,\ldots,6$$, we prove `DifferentiableOn ℂ (Φⱼ' r) ℍ₀` and `ContDiffOn ℂ ∞ (Φⱼ' r) ℍ₀` in terms of `DifferentiableOn ℂ φ₀'' ℍ₀`. We also state `ContDiffOn ℝ ∞ (Φⱼ' r) ℍ₀`.
3. In `SpherePacking.MagicFunction.a.Integrability.RealIntegrands`, we prove (real) smoothness results for `Φ₁...Φ₆` in terms of the corresponding results for `Φ₁'...Φ₆'`.
4. In `SpherePacking.MagicFunction.a.Integrability.Integrability`, we reduce the proofs that `Φ₁...Φ₆` are integrable to proofs that they have finite integrals by using smoothness to deduce `AEStronglyMeasurable`. These are left as TODOs but should be shown using the bounds computed in `SpherePacking.MagicFunction.a.IntegralEstimates`.

### DIFF
--- a/SpherePacking.lean
+++ b/SpherePacking.lean
@@ -63,6 +63,7 @@ import SpherePacking.ModularForms.PhiTransform
 import SpherePacking.ModularForms.QExpansion
 import SpherePacking.ModularForms.ResToImagAxis
 import SpherePacking.ModularForms.SlashActionAuxil
+import SpherePacking.ModularForms.ThetaDerivSlashActions
 import SpherePacking.ModularForms.clog_arg_lems
 import SpherePacking.ModularForms.csqrt
 import SpherePacking.ModularForms.equivs
--- a/SpherePacking/ModularForms/Derivative.lean
+++ b/SpherePacking/ModularForms/Derivative.lean
@@ -80,11 +80,9 @@ theorem D_add (F G : ℍ → ℂ) (hF : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) F) (
   calc
     D (F + G) z
     _ = (2 * π * I)⁻¹ * deriv ((F ∘ ofComplex) + (G ∘ ofComplex)) z := by rfl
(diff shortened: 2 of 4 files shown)


SpherePacking/MagicFunction/IntegralParametrisations.lean
```diff
@@ -44,40 +47,61 @@ end Parametrisations
 
 section UpperHalfPlane
 
--- Show that the things that go into φ₀ are in the upper half plane
+open scoped UpperHalfPlane
 
-lemma z₁_in_upper_half_plane {t : ℝ} (ht : t ∈ Ioo 0 1) : 0 < (z₁' t).im := by
-  have ht' : t ∈ Icc 0 1 := mem_Icc_of_Ioo ht
+lemma im_z₁'_pos {t : ℝ} (ht : t ∈ Ioc 0 1) : 0 < (z₁' t).im := by
```
seewoo5, line 52: Is there a reason to change from `Ioo` to `Ioc`?

thefundamentaltheor3m: I figured more generality couldn't hurt. I mean the only point we really want to exclude is $$t = 0$$ because that's when the contour touches the real axis - everywhere else, it does lie on the upper half-plane. If we ever need `Ioo` we can always use `mem_Ioc_of_Ioo`.

PR #213, opened 2025-11-30
Add proofs by Gauss in `ModularForms/Derivative.lean`

gauss-math-inc:
This PR contributes 2 proofs for previously incomplete theorems automatically generated and optimized by Gauss, Math Inc's autoformalization agent.

Theorems solved:
- `E₂_holo'`
- `antiDerPos`

---------

PR #239, opened 2025-12-15
feat(ModularForms): Monotonicity of F/G (Proposition 8.12)

cameronfreer:
## Summary

This PR proves **Blueprint Proposition 8.12**: the function `FmodGReal(t) = F(it)/G(it)` is strictly decreasing on `(0, ∞)`.

### Main results (`FG.lean`)

* `G_vanishing_order` : `G / q^(3/2) → 20480` as `im(z) → ∞`
* `D_G_div_G_tendsto` : `(D G)/G → 3/2` as `im(z) → ∞`
* `L₁₀_imag_axis_real` : `L₁,₀(it)` is real for all `t > 0`
* `L₁₀_div_FG_tendsto` : `L₁,₀(it) / (F(it)G(it)) → 1/2` as `t → ∞`
* `L₁₀_eventually_pos_imag_axis` : `L₁,₀(it) > 0` for large `t`
* `L₁₀_pos` : `L₁,₀(it) > 0` for all `t > 0` (via `antiSerreDerPos`)
* `FmodGReal_differentiableOn` : `FmodGReal` is differentiable on `(0, ∞)`
* `deriv_FmodGReal` : derivative formula `(-2π) * L₁₀(it) / G(it)²`
* `deriv_FmodGReal_neg` : `deriv FmodGReal t < 0` for all `t > 0`
* `FmodG_strictAntiOn` : `FmodGReal` is strictly decreasing on `(0, ∞)`

### Supporting infrastructure

**Derivative.lean:**
* `MDifferentiable_div` : Division of MDifferentiable functions (when denominator nonvanishing)
* `D_tendsto_zero_of_isBoundedAtImInfty` : D-derivative of bounded holomorphic function → 0 at infinity (Cauchy estimate)

**JacobiTheta.lean:**
* `differentiableAt_jacobiTheta₂_half` : Differentiability of `t ↦ jacobiTheta₂(t/2, t)`
* `Θ₂_MDifferentiable` : MDifferentiability of `Θ₂`

**FG.lean helpers:**
* `logderiv_mul_eq` : `D(f·h)/(f·h) = Df/f + Dh/h` (pointwise, where defined)
* `div_re_of_im_eq_zero` : `(a/b).re = a.re / b.re` when `b` is real-valued
* `D_cexp_div` : `D(exp(c·z))/exp(c·z) = c/(2πi)`
* `H₂_eventually_ne_zero` : Eventual nonvanishing of `H₂`
* `D_B_tendsto_zero` : Cauchy-estimate vanishing for the B polynomial
* `qexp_tendsto_of_poly_bound` : Generic q-expansion limit from polynomial coefficient bounds

### Sorry status

All new theorems are sorry-free. Pre-existing sorries remain in:
* `F_eq_FReal`, `G_eq_GReal`, `FmodG_eq_FmodGReal` (sorry'd connectors)
* `FReal_Differentiable`, `GReal_Differentiable` (future work)
* `MLDE_F`, `MLDE_G` (modular linear differential equations)
* `FmodG_rightLimitAt_zero`, `FG_inequality_1`, `FG_inequality_2` (future work)

## Dependencies

All merged: #259, #269, #270

### DIFF
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -135,16 +135,13 @@ private lemma logderiv_mul_eq (f h : ℍ → ℂ)
     (hf_md : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) f) (hh_md : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) h)
     (z : ℍ) (hf_ne : f z ≠ 0) (hh_ne : h z ≠ 0) :
     D (f * h) z / (f z * h z) = D f z / f z + D h z / h z := by
-  rw [congrFun (D_mul f h hf_md hh_md) z]
-  simp only [Pi.mul_apply, Pi.add_apply]
+  simp only [congrFun (D_mul f h hf_md hh_md) z, Pi.mul_apply, Pi.add_apply]
   field_simp [hf_ne, hh_ne]
 
-/-- `(a / b).re = a.re / b.re` when both `a` and `b` are real-valued complex numbers. -/
-private lemma div_re_of_im_eq_zero {a b : ℂ} (ha : a.im = 0) (hb : b.im = 0) :
+/-- `(a / b).re = a.re / b.re` when `b` is a real-valued complex number. -/
+private lemma div_re_of_im_eq_zero {a b : ℂ} (hb : b.im = 0) :
     (a / b).re = a.re / b.re := by
-  conv_lhs => rw [show a = ↑a.re from Complex.ext rfl (by simp [ha]),
-    show b = ↑b.re from Complex.ext rfl (by simp [hb]), ← Complex.ofReal_div]
-  exact Complex.ofReal_re _
+  rw [show b = ↑b.re from Complex.ext rfl (by simp [hb])]; exact Complex.div_ofReal_re a b.re
 
 /- Positivity of (quasi)modular forms on the imaginary axis. -/
 
@@ -155,11 +152,9 @@ Simplifies `2πi * n * z` where z=it to `-2πnt`. -/
 lemma qexp_arg_imag_axis_pnat (t : ℝ) (ht : 0 < t) (n : ℕ+) :
     2 * ↑Real.pi * Complex.I * ↑n * ↑(⟨Complex.I * t, by simp [ht]⟩ : UpperHalfPlane) =
     (-(2 * Real.pi * (n : ℝ) * t) : ℝ) := by
-  have h1 : 2 * ↑Real.pi * Complex.I * (⟨Complex.I * t, by simp [ht]⟩ : UpperHalfPlane) * n =
-      (-(2 * Real.pi * (n : ℝ) * t) : ℝ) := by
-    simpa using exp_imag_axis_arg (t := t) ht n
-  simp only [mul_assoc, mul_left_comm, mul_comm] at h1 ⊢
-  convert h1 using 2
+  have h := exp_imag_axis_arg t ht n
+  simp only [mul_assoc, mul_left_comm, mul_comm] at h ⊢
+  convert h using 2
 
 /-- Generic summability for n^a * σ_b(n) * exp(2πinz) series.
 Uses σ_b(n) ≤ n^(b+1) (sigma_bound) and a33 (a+b+1) for exponential summability. -/
@@ -882,14 +877,11 @@ theorem D_cexp_div (c : ℂ) (z : ℍ) :
     D (fun w => cexp (c * w)) z / cexp (c * z) = c / (2 * π * I) := by
   simp only [D]
   have h_deriv : deriv ((fun w : ℍ => cexp (c * w)) ∘ ⇑ofComplex) (z : ℂ) =
-      c * cexp (c * z) := by
-    have h_exp_deriv : HasDerivAt (fun w : ℂ => cexp (c * w)) (c * cexp (c * z)) (z : ℂ) :=
-      (Complex.hasDerivAt_exp (c * z)).scomp (z : ℂ)
-        (by simpa using (hasDerivAt_id (z : ℂ)).const_mul c)
-    exact ((UpperHalfPlane.eventuallyEq_coe_comp_ofComplex z.2).fun_comp
-      (fun w => cexp (c * w))).deriv_eq.trans h_exp_deriv.deriv
-  rw [h_deriv]
-  field_simp [Complex.exp_ne_zero]
+      c * cexp (c * z) :=
+    ((eventuallyEq_coe_comp_ofComplex z.2).fun_comp (fun w => cexp (c * w))).deriv_eq.trans
+      ((Complex.hasDerivAt_exp (c * (z : ℂ))).scomp (z : ℂ)
+        (by simpa using (hasDerivAt_id (z : ℂ)).const_mul c)).deriv
+  rw [h_deriv]; field_simp [Complex.exp_ne_zero]
 
 private theorem D_exp_pi_div_exp_pi (z : ℍ) :
     D (fun w => cexp (π * Complex.I * w)) z / cexp (π * Complex.I * z) = 1 / 2 := by
@@ -1037,7 +1029,7 @@ theorem L₁₀_div_FG_tendsto :
   rw [← hz] at hL hF hG
   have hFG_im : (F z * G z).im = 0 := by rw [Complex.mul_im, hF, hG]; ring
   have hFG_re : (F z * G z).re = (F z).re * (G z).re := by rw [Complex.mul_re, hF, hG]; ring
-  rw [div_re_of_im_eq_zero hL hFG_im, hFG_re]
+  rw [div_re_of_im_eq_zero hFG_im, hFG_re]
 
 theorem L₁₀_eventually_pos_imag_axis : ResToImagAxis.EventuallyPos L₁₀ := by
   refine ⟨L₁₀_imag_axis_real, ?_⟩


SpherePacking/ModularForms/Derivative.lean
```diff
@@ -320,6 +475,46 @@ example : D (E₄.toFun * E₄.toFun) = 2 * 3⁻¹ * E₄.toFun * (E₂ * E₄.t
...
+Chain rule for restriction to imaginary axis: `d/dt F(it) = -2π * (D F)(it)`.
+
+This connects the real derivative along the imaginary axis to the Serre derivative D.
+The key computation is:
+- The imaginary axis is parametrized by g(t) = I * t
+- By chain rule: d/dt F(it) = (dF/dz)(it) · (d/dt)(it) = F'(it) · I
+- Since D = (2πi)⁻¹ · d/dz, we have F' = 2πi · D F
+- So d/dt F(it) = 2πi · D F(it) · I = -2π · D F(it)
+-/
+theorem deriv_resToImagAxis_eq (F : ℍ → ℂ) (hF : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) F)
```
seewoo5: I think you can separate this theorem as a single PR (in the same file). I guess it may also shorten the proof of `antiDerPos` suggested in #213 (which is in review).

cameronfreer: Thanks, I've separated this as #249 (but not yet refactored #239 to build on top of that).

cameronfreer:
@seewoo5 I think I'm done with my initial pass on this. As noted in the PR description, it might be easiest to wait for #193, #251, and #252 to merge first. Do you have suggestions on what would be easiest to break off of this first for review after those? Thanks.

I'm not sure if #248 will need to merge first, though that might make things simpler. Also #213 is needed for the transitive dependencies of `Q_strictAntiOn` to be sorry-free, but I don't think this PR needs to wait on it.

One possibility is to go in order of files:
  | PR  | Content                                                                           | ~Lines | Dependencies             |
  |-----|-----------------------------------------------------------------------------------|--------|--------------------------|
  | A1  | Derivative.lean extensions (D_qexp, Ramanujan scaffolding)                        | ~300   | None (or perhaps #248) |
  | A2  | FG.lean extensions (Δ_fun_eq_Δ, L₁₀ def) + Monotonicity_ImagAxis.lean             | ~400   | A1                       |
  | A3  | Monotonicity_L10_SerreAndF.lean (F-side analysis)                                 | ~950   | A2                       |
  | A4  | Monotonicity_L10_GAndEventuallyPos.lean + Monotonicity.lean (G-side + final glue) | ~600   | A3                       |

Alternatively, we could divide it up by proof structure:
  | PR  | Content                                                                       | Dependencies |
  |-----|-------------------------------------------------------------------------------|--------------|
  | B1  | Imaginary axis positivity/realness (Monotonicity_ImagAxis.lean)               | FG defs      |
  | B2  | Serre derivative positivity (serre_D_L₁₀_pos_imag_axis from L10_SerreAndF)    | B1           |
  | B3  | Eventual positivity (L₁₀_eventually_pos_imag_axis from L10_GAndEventuallyPos) | B1           |
  | B4  | Final glue (Monotonicity.lean: Q, deriv_Q, Q_strictAntiOn)                    | B2, B3       |
  
I'm also happy to follow some other approach if you prefer.

cameronfreer:
awaiting-review

cameronfreer:
The build is expected to fail until #269 and #270 are merged. (Lean can't resolve the imports and so there will be "unknown module" errors.)

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,8 +882,716 @@ theorem D_F_div_F_tendsto :
...
+  refine ⟨max t₀ 1, by positivity, fun t ht => ?_⟩
+  have ht_pos : 0 < t := lt_of_lt_of_le one_pos (le_trans (le_max_right _ _) ht)
+  have hFG_pos := mul_pos (F_imag_axis_pos.2 t ht_pos) (G_imag_axis_pos.2 t ht_pos)
+  have h := mul_pos (ht₀ t (le_trans (le_max_left _ _) ht)) hFG_pos
+  rwa [div_mul_cancel₀ _ (ne_of_gt hFG_pos)] at h
+
 end AsymptoticAnalysis
 
+/- $\mathcal{L}_{1, 0}$ is eventually positive on the imaginary axis. -/
+lemma L₁₀_eventuallyPos : ResToImagAxis.EventuallyPos L₁₀ := L₁₀_eventually_pos_imag_axis
```
seewoo5: I don't think we need to redefine this.

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,8 +882,716 @@ theorem D_F_div_F_tendsto :
...
+    have : D H₄ z = D (fun w => Θ₄ w ^ 4) z := by congr 1
+    rw [this, D_pow4_eq Θ₄ hΘ₄_holo z]
+  simp_rw [h_D_H₄_pt]
+  have h_lim := (tendsto_const_nhds (x := (4 : ℂ))).mul
+    ((Θ₄_tendsto_atImInfty.pow 3).mul D_Θ₄_tendsto_zero)
+  simp only [mul_zero] at h_lim
+  exact h_lim.congr fun z => by ring
+
+/-- `D(2H₂² + 5H₂H₄ + 5H₄²) → 0` as `im(z) → ∞`. -/
+private theorem D_B_tendsto_zero :
```
seewoo5, line 933: Derivative of any holomorphic modular form should tends to zero at `ImInfty`, since they have $q$-expansion $F(z) = \sum_{n \ge n_0} a_n e^{2 \pi i n z}$ (or $e^{\pi i n z}$ if level is not 1). I'm not sure what would be the best way to prove this, but boundedness at `ImInfty` and holmorphicity seems enough to conclude this. I think this could be a separate PR. So what I expected to be true is the following (I don't think we need modularity for this):
```lean
theorem D_tendsto_zero {F : ℍ → ℂ} (hF : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) F) (hbdd : IsBoundedAtImInfty F) : 
    Tendsto (D F) atImInfty (nhds 0) := by sorry
```
Then all D (something) converges to zero can be replaced with this.

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,8 +882,716 @@ theorem D_F_div_F_tendsto :
...
+theorem Q_strictAntiOn : StrictAntiOn Q (Set.Ioi 0) := by
+  apply strictAntiOn_of_deriv_neg
+  · exact convex_Ioi 0
+  · exact Q_differentiableOn.continuousOn
+  · intro t ht
+    rw [interior_Ioi] at ht
+    exact deriv_Q_neg t ht
+
+/-- Corollary: `Q` is strictly anti-monotone (decreasing) as a function on positive reals. -/
+theorem Q_strictAnti : ∀ {t₁ t₂ : ℝ}, 0 < t₁ → t₁ < t₂ → Q t₂ < Q t₁ := by
```
seewoo5: There's already `FmodG_antitone` below, although we definitely need `StrictAntitone` for the remaining proof. Can you rename this and the above theorems? The naming change is suggested in #331 .

seewoo5 (changes requested):

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks -- a version of that worked in fd8eda573b3a79082c5c060ca4b680e3d737c862.
Should I leave it here or make it a separate PR?

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,12 +882,556 @@ theorem D_F_div_F_tendsto :
...
+
+/-!
+## Monotonicity of Q = F/G on the Imaginary Axis
+
+Proposition 8.12 from the blueprint: the function `Q(t) = F(it)/G(it)` is strictly
+decreasing on `(0, ∞)`.
+-/
+
+/-- `L₁,₀(it) > 0` for all `t > 0`. -/
+theorem L₁₀_pos_imag_axis : ResToImagAxis.Pos L₁₀ := L₁₀_pos
```
seewoo5: This also doesn't need to be re-defined.

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,12 +882,556 @@ theorem D_F_div_F_tendsto :
...
+
+/-- The function `Q(t) = Re(F(it)/G(it))` for `t > 0`. -/
+noncomputable def Q (t : ℝ) : ℝ :=
```
seewoo5: We already have `FmodGReal` above. Can you use that?

SpherePacking/ModularForms/FG.lean
cameronfreer: See 6d9419c2e4a8828e0fdbdec9bb8be97e7e5ebe2a -- is this what you had in mind?

That commit does the following:
  - FmodG_antitone → FmodG_strictAntiOn with type StrictAntiOn FmodGReal (Set.Ioi 0) (matching PR #331)
  - Removed unused Q_strictAnti (Q_strictAntiOn already covers the same fact)
  - Updated blueprint reference in modform-ineq.tex

SpherePacking/ModularForms/FG.lean
seewoo5: I think we can just keep this here - it is shorter than I thought.

SpherePacking/ModularForms/FG.lean
cameronfreer: Oops, thanks: a5790e069e7512573cc45ec05bf6fd6887cde5b6

SpherePacking/ModularForms/FG.lean
cameronfreer: See f0e0f2fc11ec4ff74761edcc1816e6a4dacba76c

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks -- see a37138e08eed7faa323d2aa23785019743ebbb79.

This replaces Q with FmodGReal throughout:
  - Deleted def Q and Q_eq_F_div_G (they duplicated FmodGReal)
  - Renamed: Q_differentiableOn → FmodGReal_differentiableOn, deriv_Q → deriv_FmodGReal, deriv_Q_neg → deriv_FmodGReal_neg
  - Folded Q_strictAntiOn into the existing FmodG_strictAntiOn (which was previously sorry'd)

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,13 +882,539 @@ theorem D_F_div_F_tendsto :
...
+    have : (fun z => 2 * H₂ z ^ 2) = (2 : ℂ) • (H₂ ^ 2) := by ext z; simp [smul_eq_mul]
+    rw [this]; exact hH₂sq.const_smul 2
+  have h2 : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) (fun z => 5 * H₂ z * H₄ z) := by
+    have : (fun z => 5 * H₂ z * H₄ z) = (5 : ℂ) • (H₂ * H₄) := by
+      ext z; simp [smul_eq_mul, mul_assoc]
+    rw [this]; exact (hH₂.mul hH₄).const_smul 5
+  have h3 : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) (fun z => 5 * H₄ z ^ 2) := by
+    have : (fun z => 5 * H₄ z ^ 2) = (5 : ℂ) • (H₄ ^ 2) := by ext z; simp [smul_eq_mul]
+    rw [this]; exact hH₄sq.const_smul 5
+  have hB_md := (h1.add h2).add h3
```
seewoo5: #282 just got merged, so I think these lines can be proved with single `fun_prop` (and similar for others).

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,13 +882,539 @@ theorem D_F_div_F_tendsto :
...
+    filter_upwards [h_ne_zero] with z hz
+    rw [h_Θ₂_fn]; exact logderiv_mul_eq f h hf_md hh_md z (hf_ne z) hz
+  have h_sum_limit : Filter.Tendsto (fun z => D f z / f z + D h z / h z) atImInfty
+      (nhds ((1 : ℂ) / 8)) := by
+    have hf_const : Filter.Tendsto (fun z => D f z / f z) atImInfty (nhds ((1 : ℂ) / 8)) := by
+      simp_rw [hf_logderiv]; exact tendsto_const_nhds
+    simpa using hf_const.add hDh_div_h_tendsto
+  exact h_sum_limit.congr' (by filter_upwards [h_logderiv_eq] with z hz; exact hz.symm)
+
+private theorem D_H₂_div_H₂_tendsto :
```
seewoo5: This is one key step for the `D_G_div_G_tendsto`, and I think there should be a way to prove this without mentioning `Θ₂` (or at least, mentioning it less). Considering the proof of `D_Θ₂_div_Θ₂_tendsto`, we may want to consider the ratio of `H₂` and `cexp (π * Complex.I * w)`. At least, it would be better to move `Θ₂_MDifferentiable` to JacobiTheta.lean.

SpherePacking/ModularForms/FG.lean
```diff
@@ -871,13 +882,539 @@ theorem D_F_div_F_tendsto :
...
+  D_tendsto_zero_of_isBoundedAtImInfty H₂_SIF_MDifferentiable isBoundedAtImInfty_H₂
+
+private theorem D_H₄_tendsto_zero :
+    Filter.Tendsto (fun z : ℍ => D H₄ z) atImInfty (nhds 0) :=
+  D_tendsto_zero_of_isBoundedAtImInfty H₄_SIF_MDifferentiable isBoundedAtImInfty_H₄
+
+/-- `D(2H₂² + 5H₂H₄ + 5H₄²) → 0` as `im(z) → ∞`, by the Cauchy estimate. -/
+private theorem D_B_tendsto_zero :
+    Filter.Tendsto (fun z : ℍ =>
+      D (fun w => 2 * H₂ w ^ 2 + 5 * H₂ w * H₄ w + 5 * H₄ w ^ 2) z)
```
seewoo5: Can you replace this and similar occurrences with point-free versions? I keep seeing this in your PRs and wonder if these are necessary or can be changed.

SpherePacking/ModularForms/FG.lean
seewoo5: Actually, `H₂_div_exp_tendsto` of #336 would be helpful.

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks for catching this. (I think the original large version of this and other PRs often had non-point-free versions and then these were only changed one small PR at a time.)

See 5e56b2f82713f414b7606fe2a7d156888cccd62b -- I think this contains all of the opportunities (the rest are `D f z / f z` etc.) but if you see others to change, please let me know.

SpherePacking/ModularForms/FG.lean
cameronfreer: > This is one key step for the `D_G_div_G_tendsto`, and I think there should be a way to prove this without mentioning `Θ₂` (or at least, mentioning it less). Considering the proof of `D_Θ₂_div_Θ₂_tendsto`, we may want to consider the ratio of `H₂` and `cexp (π * Complex.I * w)`. At least, it would be better to move `Θ₂_MDifferentiable` to JacobiTheta.lean.

See e7c6b4e: moved differentiableAt_jacobiTheta₂_half and Θ₂_MDifferentiable from FG.lean to JacobiTheta.lean (in the H_MDifferentiable section).

(No import cycle — avoided by using .mdifferentiableAt.comp τ τ.mdifferentiable_coe instead of DifferentiableAt_MDifferentiableAt from Derivative.lean.)

cameronfreer: > Actually, `H₂_div_exp_tendsto` of #336 would be helpful.

See 7039ff385c3d30a55edfa5795fd5ed9d85adc3b2 -- that shortened things a lot.

SpherePacking/ModularForms/FG.lean
cameronfreer: That worked in `D_B_tendsto_zero` and `D_G_div_G_tendsto`; see 5af4d9c58fca9d8e386c0037e0fb2a4cb5751d8a.

`hf_md` was harder as `fun_prop` wouldn't directly bridge `Complex.exp` → `MDifferentiable` (no composition rule).

cameronfreer:
@seewoo5 I think I made all the changes above, but please let me know if there's more.

I also extracted `MDifferentiable_div` to simplify `hh_md` -- it doesn't save tons of lines but makes things more modular and might be useful elsewhere? See [83e74d8](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/239/commits/83e74d8a28a6c22037eb6d420c13b291a4fa1cc6).

SpherePacking/ModularForms/FG.lean
```diff
@@ -850,13 +847,295 @@ theorem D_F_div_F_tendsto :
...
+  D_tendsto_zero_of_isBoundedAtImInfty H₄_SIF_MDifferentiable isBoundedAtImInfty_H₄
+
+/-- `D(2H₂² + 5H₂H₄ + 5H₄²) → 0` as `im(z) → ∞`, by the Cauchy estimate. -/
+private theorem D_B_tendsto_zero :
+    Filter.Tendsto (D (fun w => 2 * H₂ w ^ 2 + 5 * H₂ w * H₄ w + 5 * H₄ w ^ 2))
+      atImInfty (nhds 0) := by
+  have hB_md : MDifferentiable 𝓘(ℂ) 𝓘(ℂ)
+      (fun z => 2 * H₂ z ^ 2 + 5 * H₂ z * H₄ z + 5 * H₄ z ^ 2) := by
+    have : (fun z => 2 * H₂ z ^ 2 + 5 * H₂ z * H₄ z + 5 * H₄ z ^ 2) =
+        (2 : ℂ) • H₂ ^ 2 + ((5 : ℂ) • (H₂ * H₄) + (5 : ℂ) • H₄ ^ 2) := by
```
seewoo5: Is it possible to define `B` as this (`(2 : ℂ) • H₂ ^ 2 + ((5 : ℂ) • (H₂ * H₄) + (5 : ℂ) • H₄ ^ 2)`) instead of `fun z => 2 * H₂ z ^ 2 + 5 * H₂ z * H₄ z + 5 * H₄ z ^ 2`) so that we don't need these extra steps showing two things are the same? (More precisely, `B` in `D_G_div_G_tendsto`)

cameronfreer: Thanks, that worked -- see a56bf9e9f93318da2cb7020bc563dcace87797d0.

  1. A and B now use Pi notation (lines 955-956): let A := H₂ ^ 3 and let B := (2 : ℂ) • H₂ ^ 2 + (5 : ℂ) • H₂ * H₄ + (5 : ℂ) • H₄ ^ 2
  2. hG_eq uses G_eq directly (line 957): have hG_eq : G = A * B := G_eq — no more fun z => rfl
  3. hA and hB are one-line fun_prop (lines 958-959): no more rewriting to Pi form first
  4. Removed hH₂ local binding — using H₂_MDifferentiable directly
  5. D_B_tendsto_zero also uses Pi notation (lines 941-950): matching B's definition

cameronfreer:
also golfed a bit more in 93bff09:


  1. logderiv_mul_eq (line 138): Merged rw + simp only into single simp only (-1 line)
  2. div_re_of_im_eq_zero (line 142-144): Used Complex.div_ofReal_re instead of manually converting both a and b to ofReal. Also removed unnecessary ha
  hypothesis — only hb is needed (-2 lines, -1 param)
  3. qexp_arg_imag_axis_pnat (line 152-157): Dropped explicit type annotation for intermediate have, using exp_imag_axis_arg t ht n directly (-2 lines)
  4. D_cexp_div (line 876-884): Collapsed two nested have blocks into single term-mode .trans chain, combined rw + field_simp on one line (-3 lines)

PR #244, opened 2025-12-16
chore: bump to `v4.27.0-rc1`

pitmonticone:


lakefile.toml
```diff
@@ -11,7 +11,6 @@ linter.flexible = false
 [[require]]
 name = "mathlib"
 git = "https://github.com/leanprover-community/mathlib4.git"
-rev = "v4.27.0-rc1"
```
dwrensha: Why did you delete this line? I'm now seeing the warning:
```
warning: manifest out of date: git revision of dependency 'mathlib' changed; use `lake update mathlib` to update it
```
because the manifest still has https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/blob/0d4e4254d670822918564403ae219bde06ecab36/lake-manifest.json#L21

pitmonticone: It seems to be the usual practice in this repo.

pitmonticone: I tend to leave it in mine, but in some projects opinions differ.

dwrensha: I agree that it's okay to leave it out in the `lakefile.toml`.
But if you don't call `lake update mathlib` after updating `lakefile.toml`, then `lake` considers `lake-manifest.json` to be out-of-date.

Fixed in #246.

pitmonticone: We need to but we don't want to update Mathlib once more after the toolchain bump. We want the exact tagged version.

dwrensha: #246 keeps the same pinned `rev`. It just nulls the `inputRev` field.

pitmonticone: Yes, that's why I approved it ;)

PR #245, opened 2025-12-16
chore: remove upstreamed `summable_hammerTime` lemma

dwrensha:
`summable_hammerTime` was upstreamed as [`summable_inv_of_isBigO_rpow_inv`](https://github.com/leanprover-community/mathlib4/blob/8ee8525c1e1c3a2794c2bfa32465f220ecf0d1a4/Mathlib/NumberTheory/ModularForms/EisensteinSeries/Summable.lean#L206-L213) in https://github.com/leanprover-community/mathlib4/pull/26005.

### DIFF
--- a/SpherePacking/ModularForms/summable_lems.lean
+++ b/SpherePacking/ModularForms/summable_lems.lean
@@ -479,17 +479,6 @@ private lemma aux (a b c : ℝ) (ha : 0 < a) (hb : 0 < b) (hc : 0 < c) : a⁻¹ 
     simp only [one_div]
     apply mul_pos hc (inv_pos.mpr hb)
 
-lemma summable_hammerTime {α : Type} [NormedField α] [CompleteSpace α] (f : ℤ → α) (a : ℝ)
-    (hab : 1 < a)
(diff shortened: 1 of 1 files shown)


SpherePacking/ModularForms/summable_lems.lean
```diff
@@ -506,7 +495,7 @@ theorem summable_diff_denom (z : ℍ) (i : ℤ) :
     enter [1]
     ext m
     rw [← mul_inv]
-  apply summable_hammerTime _ 2 (by norm_num)
+  apply summable_inv_of_isBigO_rpow_inv one_lt_two
```
dwrensha, line 498: `a` is now an implicit parameter, so it would not work to do just
```lean
  apply summable_inv_of_isBigO_rpow_inv (by norm_num)
```

We could do
```lean
  apply summable_inv_of_isBigO_rpow_inv (a := 2) (by norm_num)
```
but I thought that `one_lt_two` is nicer.

b-mehta (approved):
Thanks!

PR #269, opened 2026-01-05
feat(ModularForms): add imaginary axis properties for monotonicity

cameronfreer:
## Summary

Adds imaginary axis properties for theta functions and F/G - foundational lemmas for proving Q = F/G is monotonic (Blueprint Proposition 8.12).

### Changes

**Distributed to existing files** (per reviewer feedback, no new file created):

**JacobiTheta.lean:**
- Helper lemmas: `neg_one_zpow_im_eq_zero`, `Complex.im_tsum_eq_zero_of_im_eq_zero`
- `Θ₂_term_imag_axis_real`, `Θ₂_imag_axis_real`, `Θ₂_term_imag_axis_re`, `Θ₂_term_imag_axis_re_pos`, `Θ₂_imag_axis_re_pos`
- `Θ₄_term_imag_axis_real`, `Θ₄_imag_axis_real`
- `H₂_imag_axis_real`, `H₂_imag_axis_pos`
- `H₄_imag_axis_real`, `H₄_imag_axis_pos`

**FG.lean:**
- `MDifferentiable.pow` helper (until upstreamed to mathlib)
- `G_holo` simplified using the helper
- `G_imag_axis_real`, `G_imag_axis_pos` (proved using `H₂_imag_axis_pos`, `H₄_imag_axis_pos`)
- `F_eq_nine_DE₄_sq`: identity `F = 9 * (D E₄)²`
- `F_imag_axis_real`: proved via realness of E₂/E₄/E₆
- **Direct q-expansion approach** for DE₄:
  - `sigma_qexp_summable_generic`: generic summability for n^a * σ_b(n) * qⁿ series (uses `sigma_bound` + `a33`)
  - `sigma3_qexp_summable`, `sigma3_qexp_deriv_bound`: infrastructure for σ₃ series (via generic lemma)
  - `E₄_sigma_qexp`: E₄ as explicit tsum
  - `DE₄_qexp`: D E₄ = 240 * ∑ n*σ₃(n)*qⁿ (via `D_qexp_tsum_pnat`)
  - `DE₄_imag_axis_pos`: positivity via q-expansion (each term positive)
- `F_imag_axis_pos`: proved using `F_eq_nine_DE₄_sq` and `DE₄_imag_axis_pos`
- `Δ_fun_imag_axis_pos`: positivity via `Delta_imag_axis_pos` and `CuspForm_to_ModularForm_Fun_coe`

**FG.lean (negDE₂ chain):**
- `E₂_sigma_qexp`: E₂ q-expansion in σ₁ form (wrapper around `E₂_eq` + `tsum_eq_tsum_sigma`)
- `sigma1_qexp_summable`: summability via `sigma_qexp_summable_generic`
- `sigma1_qexp_deriv_bound`: derivative bound via `iter_deriv_comp_bound3`
- `negDE₂_qexp`: q-expansion via `D_qexp_tsum_pnat` termwise differentiation
- `negDE₂_summable`, `negDE₂_imag_axis_real`, `negDE₂_imag_axis_re_pos`, `negDE₂_imag_axis_pos`

### Key design decision

**DE₄_imag_axis_pos proved directly via q-expansion**, bypassing the `E₂_mul_E₄_sub_E₆` lemma (which has a sorry). This avoids a potential circular dependency with Ramanujan's identity.

### Style changes (per reviewer feedback)

- Renamed `D_E₄` to `DE₄` (removed underscore)
- Renamed `_im` suffix to `_real` for consistency with `ResToImagAxis.Real`
- Moved `ImagAxisProperties` section earlier in FG.lean
- Golfed proofs: removed ~170 lines of redundant code

### PR Chain

This is part of splitting #239 into independent PRs:

| PR | Content | Status |
|----|---------|--------|
| #278 | Eisenstein imaginary axis realness | ✅ Merged |
| **This PR (#269)** | JacobiTheta + FG imaginary axis properties | [now merged] |
| #270 | L₁₀ Serre derivative analysis | Depends on this PR |
| #239 | G-side + Main monotonicity theorem | Depends on all above |

### Test Plan

- [x] `lake build` succeeds
- [x] `G_imag_axis_pos` and `F_imag_axis_pos` fully proved
- [x] `DE₄_imag_axis_pos` proved via direct q-expansion (no sorry dependencies)
- [x] `Δ_fun_imag_axis_pos` filled (via `Delta_imag_axis_pos`)
- [x] `negDE₂_imag_axis_pos` filled (uses `negDE₂_qexp`)

### DIFF
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -143,16 +143,6 @@ lemma Δ_fun_imag_axis_pos : ResToImagAxis.Pos Δ_fun := by
   rw [hΔ_eq]
   exact Delta_imag_axis_pos
 
-/-- `E₂ * E₄ - E₆` is real on the imaginary axis (E₂, E₄, E₆ are all real). -/
-lemma E₂_mul_E₄_sub_E₆_imag_axis_real : ResToImagAxis.Real (E₂ * E₄.toFun - E₆.toFun) := by
-  have hProd : ResToImagAxis.Real (E₂ * E₄.toFun) :=
-    ResToImagAxis.Real.mul E₂_imag_axis_real E₄_imag_axis_real
-  have hNeg : ResToImagAxis.Real ((-1 : ℝ) • E₆.toFun) :=
-    ResToImagAxis.Real.smul E₆_imag_axis_real
-  have hEq : E₂ * E₄.toFun - E₆.toFun = E₂ * E₄.toFun + (-1 : ℝ) • E₆.toFun := by
-    ext z; simp [sub_eq_add_neg]
-  simpa [hEq] using ResToImagAxis.Real.add hProd hNeg
-
 /-- The q-expansion exponent argument on imaginary axis z=it with ℕ+ index.
 Simplifies `2πi * n * z` where z=it to `-2πnt`. -/
 lemma qexp_arg_imag_axis_pnat (t : ℝ) (ht : 0 < t) (n : ℕ+) :


SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
```diff
@@ -0,0 +1,880 @@
...
+      push_cast; ring
+    rw [hcast]
+    exact Complex.ofReal_im _
+  have h2 : (Complex.I ^ k : ℂ).im = 0 := I_pow_even_real k hk
+  have heq : (-2 * Real.pi : ℂ) ^ k = (-(2 * Real.pi)) ^ k := by ring
+  rw [heq]
+  simp [Complex.mul_im, h1, h2]
+
+/-- Each term Θ₂_term n (I*t) has zero imaginary part for t > 0. -/
+lemma Θ₂_term_imag_axis_im (n : ℤ) (t : ℝ) (ht : 0 < t) :
```
seewoo5:
```suggestion
lemma Θ₂_term_imag_axis_real (n : ℤ) (t : ℝ) (ht : 0 < t) :
```
Might be a better name considering that we use `ResToImagAxis.Real` instead of `ResToImagAxis.Imag`.

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
```diff
@@ -0,0 +1,880 @@
...
+    have hI : Complex.I ^ 2 = -1 := I_sq
+    push_cast
+    ring_nf
+    simp only [hI]
+    ring
+  rw [hexpr]
+  exact exp_ofReal_im _
+
+/-- Θ₂(I*t) has zero imaginary part for t > 0. -/
+lemma Θ₂_imag_axis_im (t : ℝ) (ht : 0 < t) :
```
seewoo5: Same here. Also, this and above can be located under `JacobiTheta.lean` (I hope there's no cycle imports for that)

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
```diff
@@ -0,0 +1,880 @@
...
+`G` is holomorphic on the upper half-plane.
+Blueprint: G = H₂³ (2H₂² + 5H₂H₄ + 5H₄²) is holomorphic since H₂ and H₄ are
+holomorphic.
+-/
+theorem G_holo : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) G := by
+  -- G = H₂³ * (2H₂² + 5H₂H₄ + 5H₄²), composition of holomorphic functions
+  have hH₂ : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) H₂ := H₂_SIF_MDifferentiable
+  have hH₄ : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) H₄ := H₄_SIF_MDifferentiable
+  unfold G
+  have hH₂_sq' : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) (fun z => H₂ z ^ 2) := by
```
seewoo5: Maybe you can replace everything here with point-free versions (which make things shorter?).

This and other PRs actually motivates me to upstream something like `MDifferentiable.sq` in mathlib... I'll let you know when it happens. Then you should be able to `H₂_SIF_MDifferentiable.sq`. Much better is to make MDifferentiable version of [`fun_prop`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Tactic/FunProp.html) (someone told me about this, haven't tried yet).

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
```diff
@@ -0,0 +1,880 @@
...
+  intro t ht
+  simp only [Function.resToImagAxis, ResToImagAxis, ht, ↓reduceDIte, H₂]
+  -- H₂ = Θ₂^4, and Θ₂(I*t) has zero imaginary part,
+  -- so H₂(I*t) = Θ₂(I*t)^4 has zero imaginary part
+  have hΘ₂_im := Θ₂_imag_axis_im t ht
+  exact Complex.im_pow_eq_zero_of_im_eq_zero hΘ₂_im 4
+
+/-- Each term Θ₂_term n (I*t) has positive real part equal to exp(-π(n+1/2)²t) for t > 0. -/
+lemma Θ₂_term_imag_axis_re (n : ℤ) (t : ℝ) (ht : 0 < t) :
+    (Θ₂_term n ⟨Complex.I * t, by simp [ht]⟩).re =
```
seewoo5:
```suggestion
    (Θ₂_term n ⟨I * t, by simp [ht]⟩).re =
```
Since you opened `Complex` (and similar for others).

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
```diff
@@ -0,0 +1,880 @@
...
+-/
+theorem G_imag_axis_real : ResToImagAxis.Real G :=
+  G_imag_axis_pos.1
+
+/-!
+### Helper lemmas for Eisenstein series on imaginary axis
+-/
+
+/-- `E₄(it)` is real for all `t > 0`. -/
+theorem E₄_imag_axis_real : ResToImagAxis.Real E₄.toFun := by
```
seewoo5: For Eisenstein related things, would it be possible to locate these under `SpherePacking/ModularForms/Eisenstein.lean`?

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
cameronfreer: Thanks, I've renamed the suffixes in the various files (in the same commits where they were moved to other files).

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
cameronfreer: I fixed this (across a few commits) for results in `FG.lean` or `JacobiTheta.lean` where we have `open UpperHalfPlane hiding I`, but I wasn't sure how best to do this in `Eisenstein.lean` where both `Complex` and `UpperHalfPlane` are open and both `I` are needed at various points (and I didn't see a good way to change which one was in effect for a given section without reorganizing a lot). If you see a good way to do this in `Eisenstein.lean` I'm happy to try it there too.

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
cameronfreer: I've tried a version of this in ee9bc5611491175cdfc84ed5fcf158b4b46ab5d8 -- is that roughly what you have in mind for now?

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
seewoo5: Then let's keep `Complex.I` in Eisenstein; this is not super important.

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
seewoo5: There are still a lot of non point-free things. While I try to fix this myself, I found that we have `H₂_SIF_MDifferentiable` but not `H₂_MDifferentiable`, which is a bit annoying. So I make PR #273 adding this, then you should be able to do `MDifferentiable 𝓘(ℂ) 𝓘(ℂ) ((2 : ℂ) • H₂ ^ 2)` using `H₂_MDifferentiable`.

In fact, I believe that the best way is to make `MDifferentiable` FunProp-able and simply do `by fun_prop` for everything similar to this. I made [upstream PR](https://github.com/leanprover-community/mathlib4/pull/33808#issuecomment-3730493616) for this as the first step.

~If you don't mind, I'd upsteam `MDifferentiable.pow` too and add `fun_prop` tag to it, so that we can replace all these long proofs as one-liner once all these merged to mathlib and version is updated.~ made [PR](https://github.com/leanprover-community/mathlib4/pull/33814)

SpherePacking/ModularForms/FG.lean
```diff
@@ -179,3 +198,212 @@ theorem FG_inequality_1 {t : ℝ} (ht : 0 < t) :
...
+/--
+`G(it) > 0` for all `t > 0`.
+Blueprint: Lemma 8.6 - follows from H₂(it) > 0 and H₄(it) > 0.
+G = H₂³ (2H₂² + 5H₂H₄ + 5H₄²) is positive since all factors are positive.
+-/
+theorem G_imag_axis_pos : ResToImagAxis.Pos G := by
+  unfold G
+  have hH₂ : ResToImagAxis.Pos H₂ := H₂_imag_axis_pos
+  have hH₄ : ResToImagAxis.Pos H₄ := H₄_imag_axis_pos
+  have hH₂_sq : ResToImagAxis.Pos (fun z : ℍ => H₂ z ^ 2) := by
```
seewoo5, line 616:
```suggestion
  have hH₂_sq : ResToImagAxis.Pos (H₂ ^ 2) := by
```
and similar for others

SpherePacking/ModularForms/Eisenstein.lean
```diff
@@ -799,3 +799,246 @@ theorem E₂_mul_E₄_sub_E₆ (z : ℍ) :
...
+    rw [h]
+    exact ofReal_im _
+  have hinv_zeta_im : (1 / riemannZeta 4).im = 0 := by
+    rw [Complex.div_im, Complex.one_im, Complex.one_re, hzeta_im]
+    ring
+  simp only [Complex.mul_im, Complex.div_im, hinv_zeta_im, hsum_im, hpow_im, hfact_im]
+  ring
+
+/-- `E₆(it)` is real for all `t > 0`. -/
+theorem E₆_imag_axis_real : ResToImagAxis.Real E₆.toFun := by
```
seewoo5: For me, proofs of `E₄_imag_axis_real` and `E₆_imag_axis_real` looks quite identical, and it seems possible to prove that $E_{2k}$ is real for general $k \ge 2$ with essentially same method (probably need to consider $E_2$ separately). Could you try this, and if it works, could make it as a separate PR? So the new PR will be about real-ness of Eisenstein series.

SpherePacking/ModularForms/Monotonicity_ImagAxis.lean
grunweg: Coming here from the mathlib PR: do you know about the new `MDiff` notation? If you import `Mathlib.Geometry.Manifold.Notation`, you should be able to write `MDiff ((2 : ℂ) • H₂ ^ 2)` instead. (Sometimes, you need to add a type ascription, when Lean could previously use the model with corners you gave to figure out the domain and codomain.) Let me know if this fails; that may be a bug I'd like to fix.

seewoo5: @grunweg At least I didn't know that! Thank you for letting me know.

SpherePacking/ModularForms/FG.lean
seewoo5: Actually, I'm planning to add `fun_prop` to `ResToImagAxis`-related theorems, where the eventual goal is to give short proof as
```lean
theorem G_imag_axis_pos : ResToImagAxis.Pos G := by
  unfold G
  fun_prop
```
(see #276) After I checked locally, this seems possible after this PR get merged and have sorry-free proofs of `H₄_imag_axis_real` and similar theorems. So style change into point-free versions may not be necessary; at least this checks the statements are correct.

SpherePacking/ModularForms/Eisenstein.lean
cameronfreer: > For me, proofs of `E₄_imag_axis_real` and `E₆_imag_axis_real` looks quite identical, and it seems possible to prove that E 2 k is real for general k ≥ 2 with essentially same method (probably need to consider E 2 separately). Could you try this, and if it works, could make it as a separate PR? So the new PR will be about real-ness of Eisenstein series.

That worked, thanks. See #278 (which I think can be reviewed independently).

Also just to confirm -- is it correct that some `E*_imag_axis_real` use `.toFun` and others do not (currently E2 does not, E4 and E6 do) based on the defs in `E2.lean` and the way they are used in `FG.lean`, or should this choice be uniform?

SpherePacking/ModularForms/FG.lean
cameronfreer: > Actually, I'm planning to add `fun_prop` to `ResToImagAxis`-related theorems, where the eventual goal is to give short proof as
> 
> ```lean
> theorem G_imag_axis_pos : ResToImagAxis.Pos G := by
>   unfold G
>   fun_prop
> ```
> 
> (see #276) After I checked locally, this seems possible after this PR get merged and have sorry-free proofs of `H₄_imag_axis_real` and similar theorems. So style change into point-free versions may not be necessary; at least this checks the statements are correct.

Sounds good, thanks. 

Based on the description of #276, it looks like the plan is to finish this #269 first and then revise the proofs in a subsequent PR after #276 is merged (rather than to merge #275/#276 first and then revise this #269)?

seewoo5: Yes exactly, I want to handle #276 in a subsequent PR.

SpherePacking/ModularForms/Eisenstein.lean
seewoo5: The main reason is that $E_2$ is simply a function, while other $E_k$'s are `ModularForm` with slash-invariance and bounded at $i\infty$. At least in this sphere packing project, we are not planning to formalize something like `QuasiModularForm`; if we do, then it would be more uniform but needs considerably more work.

SpherePacking/ModularForms/FG.lean
```diff
@@ -178,3 +197,212 @@ theorem FG_inequality_1 {t : ℝ} (ht : 0 < t) :
...
+      ext z
+      simp [sub_eq_add_neg]
+    simpa [hEq] using ResToImagAxis.Real.add hProd hNeg
+  simpa [pow_two] using ResToImagAxis.Real.mul hSub hSub
+
+/--
+`F(it) > 0` for all `t > 0`.
+Blueprint: Follows from the q-expansion (E₂E₄ - E₆ = 720 * ...) and positivity.
+-/
+theorem F_imag_axis_pos : ResToImagAxis.Pos F := by
```
seewoo5, line 675: There are already `F_pos` and `G_pos` stated with sorry above in the same file. But I actually want to keep this naming and replace `F_pos`, `G_pos` with these; which is somehow reflected in the draft PR #279, but I think we can do it here.

Also, the intended proof is to prove `negDE₂_pos` first (which might be renamed by `negDE₂_imag_axis_pos` and prove this using `mul` and `smul` of `ResToImagAxis.Pos` (note that `F` is a constant multiple of the square of`(- D E₂)`. Could you separate this proof following the idea? (Eventually, `F_imag_axis_pos` will be also proved by `fun_prop` and the identity in #279 )

SpherePacking/ModularForms/FG.lean
seewoo5: The subsequent PR is #279; See also other comment.

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks, makes sense, except that I see `F = 9 * (D E₄)²` in the blueprint -- is this what you meant instead? Sorry if I'm missing something.

seewoo5: Yes exactly. Also positivity of negDE2 is used later for the monotonicity (it is X42 up to constant, which is replaced with negDE2)

cameronfreer: Thanks. I think I've renamed and restructured things as you suggested, but please have a look.

I'm also not sure if `E₂_mul_E₄_sub_E₆_imag_axis_pos` belongs here, but is something like that helpful?

SpherePacking/ModularForms/FG.lean
```diff
@@ -178,3 +314,92 @@ theorem FG_inequality_1 {t : ℝ} (ht : 0 < t) :
...
+
+open UpperHalfPlane hiding I
+open Complex
+
+/--
+`G(it) > 0` for all `t > 0`.
+Blueprint: Lemma 8.6 - follows from H₂(it) > 0 and H₄(it) > 0.
+G = H₂³ (2H₂² + 5H₂H₄ + 5H₄²) is positive since all factors are positive.
+-/
+theorem G_imag_axis_pos : ResToImagAxis.Pos G := by
```
seewoo5, line 612: This and below theorems should be moved to somewhere above; the mai goal of this file is to prove the two inequalities using these

cameronfreer: moved in 5e687fda8b0bac9f9fa596e811b8cff48f9c17d3

SpherePacking/ModularForms/FG.lean
cameronfreer: I'm partway through proving `negDE₂_imag_axis_pos` but I now see that I possibly misunderstood -- probably I should move this and its dependencies to a new PR (and maybe another file)?

SpherePacking/ModularForms/FG.lean
```diff
@@ -91,21 +139,533 @@ Modular linear differential equation satisfied by $G$.
...
+
+/-- The real part of (E₂*E₄ - E₆)(it) is positive for t > 0. -/
+lemma E₂_mul_E₄_sub_E₆_imag_axis_re_pos (t : ℝ) (ht : 0 < t) :
+    0 < ((E₂ * E₄.toFun - E₆.toFun).resToImagAxis t).re := by
+  -- Unfold ResToImagAxis
+  simp only [Function.resToImagAxis, ResToImagAxis, ht, ↓reduceDIte]
+  simp only [Pi.mul_apply, Pi.sub_apply, ModularForm.toFun_eq_coe]
+  -- From E₂_mul_E₄_sub_E₆: E₂*E₄ - E₆ = 720 * ∑ n*σ₃(n)*q^n
+  set z : UpperHalfPlane := ⟨Complex.I * t, by simp [ht]⟩ with hz
+  have hqexp := E₂_mul_E₄_sub_E₆ z
```
seewoo5: There's an issue with this; `E₂_mul_E₄_sub_E₆` would be proved by using Ramanujan's identity for E_4 and the q-expansion of E_4. So if you use `E₂_mul_E₄_sub_E₆` in this proof, I think there could be some circular argument. 

So let's not use it. Instead, use q-expansion of E_4 and its derivative to conclude that `D E₄` is positive. Then we  use Ramanujan to proceed the remaining parts. So the intermediate theorem will be just `D_E₄_imag_axis_pos` you proved below (I prefer `DE₄` though). Argument is essentially the same as you wrote, just avoid using `E₂_mul_E₄_sub_E₆` that hasn't been proved yet (I have a vague idea where we may not need it at all in this project).

SpherePacking/ModularForms/FG.lean
```diff
@@ -91,21 +139,533 @@ Modular linear differential equation satisfied by $G$.
...
+  simp only [Complex.mul_re, hn_re, hn_im, hσ_re, hσ_im, hexp_im, hexp_re, mul_zero, sub_zero]
+  apply mul_pos
+  · apply mul_pos
+    · exact_mod_cast n.pos
+    · have := ArithmeticFunction.sigma_pos 1 (n : ℕ) n.ne_zero
+      exact_mod_cast this
+  · exact Real.exp_pos _
+
+/-- `negDE₂` is real on the imaginary axis. -/
+lemma negDE₂_imag_axis_real : ResToImagAxis.Real negDE₂ := by
```
seewoo5, line 553: In future, I think this can be shortened by proving something like
```lean
theorem {F : ℍ → ℂ} (hF : ResToImagAxis.Real F) (hF : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) F) : 
	ResToImagAxis.Real (D F) :=
```
and the realness of `E₂`. I prefer to prove the above theorem later using #276 though; so let's keep this proof for now. I belive the above theorem can be proved using `deriv_resToImagAxis_eq` - feel free to check this and make PR for it.

SpherePacking/ModularForms/FG.lean
seewoo5: You did it correctly (see other comment). I confused between  `D E₄` (for `F`) and `negDE₂` above.

SpherePacking/ModularForms/FG.lean
cameronfreer: Sounds good, thanks for catching this.

SpherePacking/ModularForms/FG.lean
cameronfreer: Sounds good -- I've made an initial version at #289

SpherePacking/ModularForms/FG.lean
cameronfreer: See 53aa9a0b2379b1fc49b5f0124f6de76aeefff801 (and subsequent commits to clean things up a bit)

SpherePacking/ModularForms/FG.lean
```diff
@@ -178,3 +869,111 @@ theorem FG_inequality_1 {t : ℝ} (ht : 0 < t) :
 theorem FG_inequality_2 {t : ℝ} (ht : 0 < t) :
     FReal t - 18 * (π ^ (-2 : ℤ)) * GReal t < 0 := by
   sorry
+
+/-!
```
seewoo5: The codes below here is not needed anymore.

cameronfreer: Great, thanks -- I was wondering if this might be the case, but first moved it to the bottom, to at least be sure that nothing else depended on it.

cameronfreer: see e84f459466438d8c9a4a69dcc010c0e22b796ef4

SpherePacking/ModularForms/FG.lean
```diff
@@ -91,21 +122,575 @@ Modular linear differential equation satisfied by $G$.
...
+        (by rw [IsCuspForm_iff_coeffZero_eq_zero]; exact E4E6_coeff_zero_eq_zero)
+      simp only [Delta_E4_E6_aux, pow_two, pow_three, DirectSum.of_mul_of, DirectSum.sub_apply,
+        Int.reduceAdd, DirectSum.of_eq_same, one_div] at hAux ⊢
+      exact congrFun hAux z
+    rw [hLHS, hRHS]; ring
+  rw [hΔ_eq]
+  exact Delta_imag_axis_pos
+
+/-- `E₂ * E₄ - E₆` is real on the imaginary axis (E₂, E₄, E₆ are all real). -/
+lemma E₂_mul_E₄_sub_E₆_imag_axis_real : ResToImagAxis.Real (E₂ * E₄.toFun - E₆.toFun) := by
```
seewoo5: We can also remove this!

seewoo5 (approved):
@cameronfreer It looks good to me, just have one last comment. There might be room for refactoring but let's leave it for later.

SpherePacking/ModularForms/FG.lean
cameronfreer: removed in 86ea7b47586e808fb04c5cdc06e4b74a833dbae6

PR #273, opened 2026-01-10
add `Hi_MDifferentiable`s

seewoo5:


SpherePacking/ModularForms/JacobiTheta.lean
```diff
@@ -345,6 +345,18 @@ lemma H₄_SIF_MDifferentiable : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) H₄_SIF :=
       simpa [Function.comp] using this)
   simpa [mdifferentiableWithinAt_univ] using hMD_within
 
+lemma H₂_MDifferentiable : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) H₂ := by
+  simpa [H₂_SIF, SlashInvariantForm.coe_mk] using
+    (H₂_SIF_MDifferentiable : MDifferentiable 𝓘(ℂ) 𝓘(ℂ) H₂_SIF)
```
cameronfreer: Not sure if this is better, but I think it works without the type annotations (and analogously with H₃ and H₄):
```suggestion
  simpa [H₂_SIF, SlashInvariantForm.coe_mk] using H₂_SIF_MDifferentiable
```

cameronfreer (approved):
LGTM

SpherePacking/ModularForms/JacobiTheta.lean
seewoo5: Thanks, I updated with these proofs.

seewoo5:
@cameronfreer Thanks! I also added the `simp` coercion lemmas, but leave the proof as you suggested (just without type annotation), mostly because of the place I put these lemmas.

PR #278, opened 2026-01-11
feat(Eisenstein): add imaginary axis realness for Eisenstein series

cameronfreer:
## Summary

- Add `E_even_imag_axis_real`: general theorem that `E_k(it)` is real for `t > 0` when `k` is even and `k ≥ 4`
- Add `E₄_imag_axis_real`, `E₆_imag_axis_real` as direct corollaries
- Add `E₂_imag_axis_real` with separate proof (different q-expansion structure)
- Add supporting lemmas: `neg_two_pi_I_pow_even_real`, `exp_imag_axis_arg`, `riemannZeta_even_im_eq_zero`

### DIFF
--- a/SpherePacking/ModularForms/Eisenstein.lean
+++ b/SpherePacking/ModularForms/Eisenstein.lean
@@ -813,7 +813,7 @@ open Complex hiding I
 /-- `(-2πi)^k` is real for even k. -/
 lemma neg_two_pi_I_pow_even_real (k : ℕ) (hk : Even k) :
     ((-2 * Real.pi * Complex.I) ^ k : ℂ).im = 0 := by
-  have h : (-2 * Real.pi * Complex.I) ^ k = ((-2 * Real.pi) ^ k : ℂ) * Complex.I ^ k := by ring
+  have h : (-2 * Real.pi * Complex.I) ^ k = (-(2 * Real.pi) : ℂ) ^ k * Complex.I ^ k := by ring
   rw [h]
   have h1 : ((-(2 * Real.pi)) ^ k : ℂ).im = 0 := by norm_cast
   have h2 : (Complex.I ^ k : ℂ).im = 0 := by
@@ -821,8 +821,6 @@ lemma neg_two_pi_I_pow_even_real (k : ℕ) (hk : Even k) :
     simp only [← two_mul, pow_mul, I_sq]
     -- (-1)^m is real: ±1
     rcases m.even_or_odd with hm | hm <;> simp [hm.neg_one_pow]
-  have heq : (-2 * Real.pi : ℂ) ^ k = (-(2 * Real.pi)) ^ k := by ring
-  rw [heq]
   simp [Complex.mul_im, h1, h2]
 
 /-- On imaginary axis z = I*t, the q-expansion exponent 2πi·n·z reduces to -(2πnt).


SpherePacking/ModularForms/Eisenstein.lean
```diff
@@ -799,3 +799,173 @@ theorem E₂_mul_E₄_sub_E₆ (z : ℍ) :
...
+
+/-- On imaginary axis z = I*t, the q-expansion exponent 2πi·n·z reduces to -(2πnt).
+This is useful for reusing the same algebraic simplification across `E₂`, `E₄`, `E₆`. -/
+lemma exp_imag_axis_arg (t : ℝ) (ht : 0 < t) (n : ℕ+) :
+    2 * Real.pi * Complex.I * (⟨Complex.I * t, by simp [ht]⟩ : ℍ) * n =
+    (-(2 * Real.pi * (n : ℝ) * t) : ℝ) := by
+  have hI : Complex.I ^ 2 = -1 := I_sq
+  push_cast
+  ring_nf
+  simp only [hI]
```
seewoo5:
```suggestion
  push_cast
  ring_nf
  simp only [I_sq]
```
Maybe?

seewoo5 (approved):

SpherePacking/ModularForms/Eisenstein.lean
cameronfreer: Thanks, that worked. Also golfed a bit more.

SpherePacking/ModularForms/Eisenstein.lean
```diff
@@ -799,3 +799,164 @@ theorem E₂_mul_E₄_sub_E₆ (z : ℍ) :
...
+lemma exp_imag_axis_arg (t : ℝ) (ht : 0 < t) (n : ℕ+) :
+    2 * Real.pi * Complex.I * (⟨Complex.I * t, by simp [ht]⟩ : ℍ) * n =
+    (-(2 * Real.pi * (n : ℝ) * t) : ℝ) := by
+  push_cast
+  ring_nf
+  simp only [I_sq]
+  ring
+
+/-- `ζ(2k)` is real for all `k ≥ 1`. -/
+lemma riemannZeta_even_im_eq_zero (k : ℕ) (hk : k ≠ 0) :
```
seewoo5, line 837: Not strongly related, but somehow it seems that "\zeta(s) is real for real s > 1" is missing in mathlib..

cameronfreer: Thanks for noticing this -- I've tried adding this as my first mathlib contribution: 
https://github.com/leanprover-community/mathlib4/pull/33936

(Not sure if we'd want to specialize that here, though, as it would require importing a lot for L-functions that we might not otherwise need.)

cameronfreer: It is now merged: https://github.com/leanprover-community/mathlib4/commit/59850a9701781fb375242f83885fb625fe281f6e

Should I eventually open a spherepacking PR to use it inline and remove our `riemannZeta_even_im_eq_zero`? It would simplify our codebase but add an `LSeries` dependency (though maybe this doesn't matter much with the new module system?). In any case I suppose we need to wait until we bump mathlib (or we could put the new proof in ForMathlib until then).

seewoo5: If importing the corresponding file doesn't take significantly more time, then I think it is good to do. But yes, we need to wait until next bump.

cameronfreer: Sounds good -- tracked in #318.

PR #275, opened 2026-01-10
`ResToImagAxis.Real.pow` and `ResToImagAxis.Pos.pow`

seewoo5:
More helper theorems for `ResToImagAxis`: const, zero, one, neg, pow.

PR #279, opened 2026-01-12
Register `ResImagAxis` properties to `fun_prop` & simplify proofs

### LINKED ISSUE #276: `fun_prop` for `ResToImagAxis.Real/Pos/EventuallyPos`
Add `fun_prop` tags to the theorems on `ResToImagAxis.Real/Pos/EventuallyPos`. Then after #269 merged and we have sorry-free proofs of `H₂_imag_axis_real` and `H₄_imag_axis_real` (and similar theorems for positivity), then this will allow us to do
```lean
theorem G_imag_axis_real : ResToImagAxis.Real G := by
  unfold G
  fun_prop

theorem G_imag_axis_pos : ResToImagAxis.Pos G := by
  unfold G
  fun_prop (disch := positivity)
```

Depends on #275

seewoo5:
Register `ResToImagAxis` to `fun_prop`. Also, change definition of `G` slightly as
```lean
noncomputable def G := H₂ ^ 3 * ((2 : ℝ) • H₂ ^ 2 + (5 : ℝ) • H₂ * H₄ + (5 : ℝ) • H₄ ^ 2)
```
to match with signatures in new theorems of #275 

Depends on
 
- #269 
- #275 
- #278

Note that the definition of `G` is also slightly changed, to match with the `smul` theorems.

TODO:
- [x] Replace proofs of `F_imag_axis_real` and `F_imag_axis_pos` with `fun_prop`

Closes #276

### DIFF
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -212,69 +212,49 @@ lemma sigma1_qexp_summable (z : UpperHalfPlane) :
   simp only [pow_zero, one_mul] at h
   exact h
 
+/-- Generic derivative bound for σ_k q-series on compact sets.
+Uses σ_k(n) ≤ n^(k+1) (sigma_bound) and iter_deriv_comp_bound3 for exponential decay. -/
+lemma sigma_qexp_deriv_bound_generic (k : ℕ) :
+    ∀ K : Set ℂ, K ⊆ {w : ℂ | 0 < w.im} → IsCompact K →
+      ∃ u : ℕ+ → ℝ, Summable u ∧ ∀ (n : ℕ+) (z : K),
+        ‖(ArithmeticFunction.sigma k n : ℂ) * (2 * Real.pi * Complex.I * n) *
+          Complex.exp (2 * Real.pi * Complex.I * n * z.1)‖ ≤ u n := by
+  intro K hK hKc
+  obtain ⟨u₀, hu₀_sum, hu₀_bound⟩ := iter_deriv_comp_bound3 K hK hKc (k + 2)
+  refine ⟨fun n => u₀ n, hu₀_sum.subtype _, fun n z => ?_⟩
+  have hpow : (2 * π * n) ^ (k + 2) * ‖Complex.exp (2 * π * Complex.I * n * z.1)‖ ≤ u₀ n := by
+    simpa [abs_of_pos Real.pi_pos] using hu₀_bound n z
+  calc ‖(ArithmeticFunction.sigma k n : ℂ) * (2 * π * Complex.I * n) *
+          Complex.exp (2 * π * Complex.I * n * z.1)‖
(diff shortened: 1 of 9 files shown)


seewoo5:
awaiting-review

SpherePacking/ModularForms/ResToImagAxis.lean
```diff
@@ -138,6 +188,42 @@ theorem ResToImagAxis.Pos.smul {F : ℍ → ℂ} {c : ℝ} (hF : ResToImagAxis.P
...
+@[fun_prop]
+theorem ResToImagAxis.Pos.smul' {F : ℍ → ℂ} {c : ℝ} (hF : ResToImagAxis.Pos F)
+    (hc : 0 < c) : ResToImagAxis.Pos (fun z => c • F z) := by
+  rw [Pos]
+  refine ⟨Real.smul hF.1, fun t ht ↦ ?_⟩
+  have hFreal := hF.1 t ht
+  have hFpos := hF.2 t ht
+  simp only [Function.resToImagAxis, ResToImagAxis, ht, ↓reduceDIte] at hFreal
+  simp only [Function.resToImagAxis, ResToImagAxis, ht, ↓reduceDIte] at hFpos
+  simp [ResToImagAxis, ht, mul_pos hc hFpos]
```
cameronfreer: Do we need both `ResToImagAxis.Pos.smul'` and `ResToImagAxis.Pos.smul` (maybe the prime version is leftover from the `hmul` -> `smul` rename)? I haven't tested this, but can `fun_prop` get one from the other as they have defeq conclusions?

SpherePacking/ModularForms/FG.lean
```diff
@@ -2,10 +2,11 @@ import Mathlib.Analysis.SpecialFunctions.Trigonometric.Basic
 import Mathlib.Order.Monotone.Defs
 
 import SpherePacking.ModularForms.Derivative
+import SpherePacking.ModularForms.Eisenstein
```
cameronfreer, line 5: I think this may be redundant now (transitively imported via `Derivative`)

SpherePacking/ModularForms/ResToImagAxis.lean
cameronfreer: Nevermind -- upon looking into it a bit more, it seems like `fun_prop` won’t eta‑reduce a goal like `ResToImagAxis.Pos (fun z => c • F z)` into `ResToImagAxis.Pos (c • F)` -- I suppose this is why they're both necessary?

cameronfreer (approved):

SpherePacking/ModularForms/ResToImagAxis.lean
seewoo5: This was indeed something I forgot to remove. I removed it back and checked `fun_prop` still works for the purposes.

PR #290, opened 2026-01-19
Serre derivative preserves realness `serre_D_real_of_real`

seewoo5:
If $F$ is real-valued, then $\partial_k F$ is also real-valued for real $k$. This will be useful to prove that certain modular forms in `FG.lean` are real-valued.

Depends on #275


cameronfreer (approved):
LGTM (once #275 merges)

thefundamentaltheor3m (approved):

PR #301, opened 2026-01-19
blueprint: remove H3 from cor:theta-pos

seewoo5:
See #299


thefundamentaltheor3m (approved):
Sorry for the late review!

PR #298, opened 2026-01-19
feat(MagicFunction): vertical-ray vanishing lemmas for double zeros

cameronfreer:
## Summary
Add VerticalVanishing.lean with vanishing lemmas for the double zeros proof.

This file proves that the vertical integrands vanish as Im(z) → ∞, which is needed for the Cauchy-Goursat deformation argument in Proposition 4.4.6 (Lemma 4.4.5).

### File Added

**VerticalVanishing.lean** (237 lines)

**Vanishing lemmas (Lemma 4.4.5)**:
- `tendsto_φ₀_integrand_atImInfty`: Base case at x = 0
  - Shows φ₀''(-1/(I·t))·(I·t)²·e^{πIr(I·t)} → 0 as t → ∞
  - Uses that this equals (-I)·verticalIntegrandX(0,r,t)

- `tendsto_φ₀_integrand_plus_one`: Shifted variant at x = 1
  - Shows φ₀''(-1/(1+I·t))·(1+I·t)²·e^{πIr(I·t)} → 0 as t → ∞
  - Bounds by topEdgeBound which vanishes for r > 2

- `tendsto_φ₀_integrand_minus_one`: Shifted variant at x = -1
  - Shows φ₀''(-1/(-1+I·t))·(-1+I·t)²·e^{πIr(I·t)} → 0 as t → ∞
  - Same technique as plus_one case

These lemmas ensure that when we close the contour by adding vertical rays to ±i∞, the contribution from these rays vanishes.

## PR Stack (merge in order)
1. [#288](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/288) ← CuspPath, RealDecay
2. [#296](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/296) ← ContourEndpoints (depends on #288)
3. [#297](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/297) ← VerticalIntegrability (depends on #296)
4. **#298** ← VerticalVanishing (this PR, depends on #296, parallel with #297)
5. [#229](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/229) ← Double zeros proof (depends on #297 + #298)

## Test plan
- [x] `lake build` passes
- [x] No sorries in VerticalVanishing.lean

PR #297, opened 2026-01-19
feat(MagicFunction): vertical-ray integrability for double zeros

cameronfreer:
## Summary
Add VerticalIntegrability.lean with integrability lemmas for the double zeros proof.

This file proves that the seven integrands from Proposition 4.4.6 are integrable on their respective domains, using the thesis bounds (Lemmas 4.4.3-4.4.4) and the shifted Möbius transform technique.

> **Depends on [#304](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/304) and [#296](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/296), which must merge first** (in that order). The φ-bounds (`φ₀_bound`, `φ₂'_bound`, `φ₄'_bound`) come from #304's standalone `PhiBounds.lean`; `ContourEndpoints` comes from #296.

### File Added

**VerticalIntegrability.lean**

**Thesis bounds (Lemmas 4.4.3-4.4.4)**:
- `norm_φ₀_I_div_t_small`: For t ∈ (0,2), ‖φ₀(i/t)‖ ≤ C₀·e^{-2π/t} (super-exponential decay)
- `norm_φ₀_I_div_t_large`: For t ≥ 2, ‖φ₀(i/t)‖ ≤ C·t⁻²·e^{2πt} (polynomial growth)

**Core integrability**:
- `integrableOn_φ₀_shifted_Möbius`: General integrability for φ₀''(-1/(a+I·t))·(a+I·t)²·e^{iπr(b+I·t)} on (1,∞)

**Seven integrability goals (Proposition 4.4.6)**:
- `integrableOn_goal1` through `integrableOn_goal7`: The specific integrands needed for the contour deformation, covering both Category A (reduces to verticalIntegrandX) and Category B (shifted Möbius at ±1)
- `goal1_eq_verticalIntegrandX`, `goal2_eq_neg_I_verticalIntegrandX`, etc.: Rewrite lemmas relating goals to verticalIntegrandX

**Helper lemmas**:
- `integrableOn_verticalIntegrandX_Ioc`: Integrability on (0,1]
- `integrableOn_verticalIntegrandX_Ioi`: Integrability on (1,∞)
- `integrableOn_Ici_of_eqOn_verticalIntegrandX`, `integrableOn_Ioi_of_eqOn_neg_I_verticalIntegrandX`: Factor out common EqOn patterns
- `integrableOn_shiftedMöbius`: Unified helper for shifted Möbius integrands

[...]

PR #296, opened 2026-01-19
feat(MagicFunction): contour endpoint bounds for double zeros

cameronfreer:
## Summary
Add ContourEndpoints.lean with bounds and integrand definitions for the double zeros proof.

This file provides the S-transform bounds (Corollaries 7.5-7.7) and defines the vertical integrand infrastructure needed for the Cauchy-Goursat deformation argument.

> **Depends on [#304](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/304), which must merge first** ([#423](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/423) has merged). This PR consumes the φ-bound lemmas (`φ₀_bound`, `φ₂'_bound`, `φ₄'_bound`) from #304's `PhiBounds.lean` (it no longer defines its own bundled `PhiBounds` structure), and the `tendsto_const_mul_rpow_mul_exp_neg_atTop` decay lemma from #423's `RealDecay.lean` (used termwise in the vertical/top-edge bound decay proofs).

### File Added

**ContourEndpoints.lean**

**φ-bound lemmas** (from [#304](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/304)): consumes `φ₀_bound`, `φ₂'_bound`, `φ₄'_bound` (Cor 7.5-7.7) — bounds on ‖φ₀‖, ‖φ₂'‖, ‖φ₄'‖ for Im(z) > 1/2.

**S-transform bounds**:
- `norm_φ₀_S_smul_le`: Bound on φ₀(S·z) using the modular S-transform
- `norm_φ₀''_I_div_t_le`: Blueprint Corollary 7.13 bound for φ₀''(I/t)

**Vertical integrand infrastructure**:
- `verticalIntegrandX x r t`: The integrand φ₀''(I/t)·(x + I·t)²·e^{iπr(x + I·t)}
- `verticalBound`: Dominating function for vertical integrands
- `integrableOn_verticalIntegrandX`: Integrability on [1,∞)
- `tendsto_verticalIntegrandX_atTop`: Vanishing as t → ∞

**Top edge infrastructure**:
- `topEdgeBound`, `topEdgeIntegrand`: For the [-1,1] × {t} horizontal segment
- `tendsto_topEdgeIntegral_zero`: Top edge contribution vanishes as t → ∞

**Cauchy-Goursat prerequisites** (new):
- `uniform_vanishing_verticalIntegrandX`: Uniform vanishing for vertical integrand
  ```
[...]

PR #261, opened 2026-01-02
feat(Integrability): add Φⱼ integrability proofs

cameronfreer:
## Summary

Completes the Φⱼ integrability proofs for the magic function contour integral components.

**Depends on: PR #260** (must be merged first)

## What This PR Adds

This PR adds ~260 lines to `Integrability.lean` providing:

### Φⱼ Integrability Proofs
- `Φ₁_integrableOn` through `Φ₆_integrableOn`: Each segment integrand is integrable on V × (parameter domain)

### Helper Lemmas
- `exp_neg_two_pi_div_mul_sq_le`: Super-exponential decay bound for cusp-approaching segments
- `Φ₆_eq_I₆_g`: Connection between Φ₆ and the g function from IntegralEstimates

## Dependency Chain

```
#260 (ProductIntegrability with Fubini swaps)
  ↓
#261 (This PR: Φⱼ integrability proofs)
```

## Files Changed

| File | Change |
|------|--------|
| `Integrability.lean` | +258 lines (main content) |
| `ContourEndpoints.lean` | +1 line (import fix) |

## Test Plan

- [x] `lake build` passes

Closes #76

PR #304, opened 2026-01-20
feat(MagicFunction): factorwise estimates for φ₀, φ₂', and φ₄'

cameronfreer:
> **This PR is NOT at the bottom of the stack. It depends on both #472 and #477, neither of which is merged yet. Merge those prerequisites first (in either order), then #304, then #473.**

## Stack and review scope

```text
#472 — coefficient definitions and growth bounds ─┐
#477 — shared pointwise q-series bounds ──────────┴─→ #304 ─→ #473
```

This branch includes both prerequisite branches. The GitHub diff against `main` therefore includes their changes: `CauchyCoeffBounds.lean` belongs to #472, and `QSeriesBounds.lean` plus the `ResToImagAxis.lean` refactor belong to #477. **The new mathematics specific to #304 is in `FourierExpansions.lean` and `PhiBounds.lean`.** After the prerequisites merge, refresh this branch from `main`; squash-merging does not automatically remove the inherited history.

## What #304 adds

- Linear-factor q-expansions and bounds for `E₄` and `E₂E₄ − E₆`, connected to the existing `DivDiscBound` machinery.
- Explicit bounds `φ₀_bound`, `φ₂'_bound`, and `φ₄'_bound`, their constants and positivity, and Big-O forms along `resToImagAxis`—as separate theorems, not a structure.

The proof bounds each factor separately in `φ₀ = (E₂E₄ − E₆) · ((E₂E₄ − E₆) / Δ)` and the analogous formulas for `φ₂'` and `φ₄'`. It does not need Cauchy-product coefficients. The exact product expansions are separate work in #473.

## Cleanup in this revision

- Uses #472's `bg_poly`, `bE₄_poly`, and `evenCoeff_poly` directly, deleting the duplicate growth lemmas. The `E₄` quotient/positivity applications now use degree `4`, not the weaker degree `5`.
- Uses #477's pointwise bounds instead of reimplementing the q-series estimate.
- Removes `bg_shift_isBigO` and `g_qexp_succ`. `norm_g_le` uses the unshifted `g_qexp_nat` and the vanishing of `bg 0`. The normalization is explicit: `B_g = exp π * ∑' m, ‖bg m‖ * exp (-π * m)`. The `exp π` compensates for the reference height `1/2`, preserving the downstream bound statements.
- Uses mathlib's `norm_mul_le_of_le` instead of the local wrapper, and trims outdated proof narration.

## Compatibility and shared support

- Uses Mathlib's `EisensteinSeries.summable_sigma_mul_cexp_pow` directly instead of `sigma3_qexp_summable`, which #410 removes. No dependency on #410 is introduced.
- An unpublished merge with #410 also builds through `PhiBounds`, checking both PRs together without adding #410 to this branch.
- Owns `tsum_fouterm_shift`, shared by `g_eq_fouterm` here and the exact product expansions in #473. #304 does not import the product-expansion file.
- Includes the latest #472 proof simplification and #477 regression tests/import cleanup. The earlier substantive deduplication was already present before this refresh.

## Verification

Full `lake build`, `lake exe mk_all --check --module`, and `git diff --check` pass. All three explicit `φ` bounds, all three Big-O results, and the rewritten `norm_g_le` use only `[propext, Classical.choice, Quot.sound]`. No new `sorry` declarations.


SpherePacking/MagicFunction/a/PhiBounds.lean
thefundamentaltheor3m, line 1: I don't really like the fact that the bounds are bundled into a `structure` here. Can we have separate theorems for each of the results? We can express the theorems in multiple forms: one with `divDiscBound` as an explicit constant and one with big O (as I think that might be easier to use later on in the argument).

thefundamentaltheor3m (changes requested):

SpherePacking/MagicFunction/a/PhiBounds.lean
cameronfreer: I've given this a try -- how does it look to you? (I also rebased on main now that #268 has landed, which allowed me to eliminate some of the `sorry` q-expansion stubs, but I have left the Fourier versions as sorries for a future PR.)

cameronfreer:
Follow-up after this lands: retire the old `norm_φ₀_le` path.

The existing `norm_φ₀_le` corollary in `PolyFourierCoeffBound.lean` still carries two sorries. One is just summability and could be discharged by the new `summable_fouterm_of_poly` pattern, but the other is the real issue: it tries to represent `(E₂ * E₄ - E₆)^2` with coefficients `c n = n * σ₃ n`. That is not the square's q-expansion; the square would require Cauchy-product coefficients. So this path should be retired rather than completed.

A separate PR can now replace it cleanly using this PR's API:

1. Delete the old `norm_φ₀_le` proof from `PolyFourierCoeffBound`.
2. Reintroduce the same existential statement in a file that imports `PhiBounds`, as the wrapper `⟨C_φ₀, C_φ₀_pos, φ₀_bound⟩`.
3. Repoint the six consumers in `IntegralEstimates/I1`–`I6` to that wrapper.

Net result: remove the stale coefficient-model sorries and route downstream through the new factorized `φ₀_bound` proof. This is independent of the #296/#297/#229 contour stack.

cameronfreer:
**Restructuring the magic-function integral PRs**

I reorganised the overlapping `MagicFunction/a` work into focused PRs.

**Existing prerequisites, refreshed**

- #304: the existing factorwise estimates for `φ₀`, `φ₂'`, and `φ₄'`. The
  latest cleanup reuses `ResToImagAxis` and deduplicates `norm_qseries_shift_le`.
- #425: reusable holomorphicity lemmas for the relevant Möbius transformations,
  now fixed against current `main`.

**New or rebuilt**

- #463: shared integrable majorants for the six contour integrands, grouped by
  contour type; proves all six `Φⱼ_integrableOn` results.
- #464: joint integrability of the two-variable integrands and justified
  interchange of the two integrations defining the `Iⱼ`.
- #465: estimates for `φ₀` near the cusp along the imaginary axis.
- #466: integrability along the vertical rays and vanishing of the closing
  horizontal edge.
- #467: reformulates `OpenRectangular` to take vanishing of the top-edge integral
  directly, while preserving the original statements as wrappers.
- #468: removes the false `PolyFourierCoeffBound.norm_φ₀_le` and proves
  `norm_φ₀''_le` from #304's valid `φ₀_bound`.

The main correctness payoff is that `PolyFourierCoeffBound.lean` is now sorry-free and `sorryAx`
disappears from all six `Φⱼ_integrableOn` theorems.

**Superseded**

- #260 and #261 → #463 and #464
- #296, #297, and #298 → #465 and #466

**Dependencies**

- #463 → #464
- #304 + #463 → #468
- #304 → #465 → #466
- #425 + #466 + #467 → focused replacement for #229

All eight current PRs are green and mergeable.

🤖 Directed by Cameron Freer, written by GPT 5.6

SpherePacking/MagicFunction/a/FourierExpansions.lean
```diff
@@ -0,0 +1,384 @@
...
+/-- ℕ-indexed `q`-coefficients of `E₄`: `1` at `0`, `240·σ₃(m)` for `m ≥ 1`. -/
+def bE₄ : ℕ → ℂ := fun m => if m = 0 then 1 else 240 * (σ 3 m : ℂ)
+
+/-- ℕ-indexed `q`-coefficients of `E₂E₄ − E₆`: `720·m·σ₃(m)` (vanishes at `0`). -/
+def bg : ℕ → ℂ := fun m => 720 * (m : ℂ) * (σ 3 m : ℂ)
+
+/-- `E₄` as an ℕ-indexed `q`-series with coefficients `bE₄`. -/
+lemma E₄_qexp_nat (z : ℍ) :
+    E₄ z = ∑' m : ℕ, bE₄ m * cexp (2 * ↑π * Complex.I * ↑m * ↑z) := by
+  have hsummable : Summable (fun m : ℕ => bE₄ m * cexp (2 * ↑π * Complex.I * ↑m * ↑z)) := by
```
CBirkbeck: isnt this alredy in some general form in mathlib already?

CBirkbeck: see here: https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/ModularForms/EisensteinSeries/QExpansion.html#EisensteinSeries.E_qExpansion_coeff

cameronfreer:
🤖Claude replying (as a top-level comment — a pending draft review blocks inline replies) to @CBirkbeck's question on `E₄_qexp_nat` ("isn't this already in some general form in mathlib?"):

Yes — and the repo already routes through it. Since #439, `E4_q_exp` in `Eisenstein.lean` is proved directly from mathlib's `EisensteinSeries.E_qExpansion_coeff` (with `E4_eq' : E₄ = ModularForm.E …` bridging our `E₄` to mathlib's normalisation), and `E₄_sigma_qexp` in `FG.lean` converts that coefficient statement to tsum form. The lemma flagged here doesn't re-prove the general fact: it only reindexes `E₄_sigma_qexp` from `ℕ+` to `ℕ` with the `bE₄` packaging that the shift bounds downstream consume. Same story for `g_qexp_nat` via the Ramanujan-identity route.

SpherePacking/MagicFunction/a/FourierExpansions.lean
cameronfreer: Yes, and the repo already makes use of this but we need some reindexing -- see https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/304#issuecomment-5497593360

PR #307, opened 2026-01-21
Compute limit $\lim_{t \to 0^+} F(it) / G(it) = 18 / \pi^2$

seewoo5:
This is mostly done by Claude Opus 4.5, and in principle non-draft, but will be splited into several PRs and styles need to be upgraded a lot. For the purpose, it would be helpful if anyone can give suggestions.

Depends on:
#308 
#323 
#324

PR #350, opened 2026-03-09
refactor: apply `tendsto_cont` + add Tendsto helper lemmas

cameronfreer:
## Summary

Applies the `tendsto_cont` tactic (from #330 and #367, now merged) across modular forms proofs, plus adds reusable helper lemmas for Tendsto reasoning.

### `tendsto_cont` call sites (12 uses across 4 files)

| File | Lemmas using `tendsto_cont` |
|------|---------------------------|
| FG.lean | `D_B_tendsto_zero`, `D_G_div_G_tendsto`, `G_div_exp_tendsto` (h_poly) |
| ThetaDerivIdentities.lean | `f₂_tendsto`, `f₄_tendsto`, `theta_g_tendsto`, `theta_h_tendsto`, `H_sum_sq_tendsto` |
| JacobiTheta.lean | `jacobi_g_tendsto`, `jacobi_f_tendsto`, `theta_prod_sq_tendsto_atImInfty` |
| RamanujanIdentities.lean | `ramanujan_E₆` (E₄² limit) |

### Tendsto helper lemmas

| Lemma | Location | Purpose |
|-------|----------|---------|
| `serre_D_tendsto_of_tendsto` | EisensteinAsymptotics.lean | Generalized: if `f → c` then `serre_D k f → -(k * c) / 12` |
| `serre_D_tendsto_zero_of_tendsto_zero` | EisensteinAsymptotics.lean | Corollary for c = 0 |

### Refactoring

- **ThetaDerivIdentities.lean:** Golfed tendsto proofs (point-free style, inlined single-use haves, simpa)
- **JacobiTheta.lean:** Replaced manual `.mul`/`.pow` chain + `norm_num` with `tendsto_cont`
- **FG.lean:** Replaced `prodMk_nhds` + `fun_prop` pattern with `tendsto_cont`
- **RamanujanIdentities.lean:** Replaced `.mul` chain with `tendsto_cont`
- **IsCuspForm.lean:** Consolidated `open scoped` declarations

## Test plan

- [x] `lake build` passes (all files)

## Follow-up

More `tendsto_cont` applications possible in #307 (Qlim) — e.g. `denominator_tendsto_at_infty`, `numerator_tendsto_at_infty` — once #367 merges.

### DIFF
--- a/SpherePacking/ForMathlib/AtImInfty.lean
+++ b/SpherePacking/ForMathlib/AtImInfty.lean
@@ -5,7 +5,7 @@ import Mathlib.Analysis.Complex.UpperHalfPlane.FunctionsBoundedAtInfty
 Probably put this at Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
 -/
 
-open UpperHalfPlane
+open UpperHalfPlane Filter Topology
 
 lemma Filter.eventually_atImInfty {p : ℍ → Prop} :
     (∀ᶠ x in atImInfty, p x) ↔ ∃ A : ℝ, ∀ z : ℍ, A ≤ z.im → p z :=
@@ -13,4 +13,12 @@ lemma Filter.eventually_atImInfty {p : ℍ → Prop} :
 
 lemma Filter.tendsto_im_atImInfty : Tendsto (fun x : ℍ ↦ x.im) atImInfty atTop :=
   tendsto_iff_comap.mpr fun ⦃_⦄ a => a
+
+/-- If f tends to c ≠ 0 at infinity, then f ≠ 0 as a function.
+
+This packages the common argument: if f = 0, then f → 0, but also f → c by hypothesis.
+By uniqueness of limits, 0 = c, contradicting c ≠ 0. -/
+lemma ne_zero_of_tendsto_ne_zero {f : ℍ → ℂ} {c : ℂ} (hc : c ≠ 0)
+    (hf : Tendsto f atImInfty (nhds c)) : f ≠ 0 := fun h =>
+  hc (tendsto_nhds_unique tendsto_const_nhds (h ▸ hf)).symm
 
--- a/SpherePacking/ForMathlib/AtImInfty.lean
+++ b/SpherePacking/ForMathlib/AtImInfty.lean
@@ -1,5 +1,9 @@
-import Mathlib.Analysis.Normed.Group.Tannery
-import Mathlib.Analysis.Complex.UpperHalfPlane.FunctionsBoundedAtInfty
+module
+
+public import Mathlib.Analysis.Normed.Group.Tannery
+public import Mathlib.Analysis.Complex.UpperHalfPlane.FunctionsBoundedAtInfty
+
+@[expose] public section
 
 /-
 Probably put this at Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
@@ -21,4 +25,3 @@ By uniqueness of limits, 0 = c, contradicting c ≠ 0. -/
 lemma ne_zero_of_tendsto_ne_zero {f : ℍ → ℂ} {c : ℂ} (hc : c ≠ 0)
     (hf : Tendsto f atImInfty (nhds c)) : f ≠ 0 := fun h =>
   hc (tendsto_nhds_unique tendsto_const_nhds (h ▸ hf)).symm
-
--- a/SpherePacking/ModularForms/EisensteinAsymptotics.lean
+++ b/SpherePacking/ModularForms/EisensteinAsymptotics.lean
@@ -216,20 +216,40 @@ def serre_DE₆_ModularForm : ModularForm (CongruenceSubgroup.Gamma 1) 8 :=
 
 /-! ## Limit of serre_D at infinity (for determining scalar) -/
 
-/-- General limit: if `f → 1` at i∞ and f is holomorphic and bounded, then `serre_D k f → -k/12`. -/
-lemma serre_D_tendsto_neg_k_div_12 (k : ℤ) (f : ℍ → ℂ)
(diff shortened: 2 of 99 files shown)


SpherePacking/ForMathlib/AtImInfty.lean
b-mehta, line 1: Perhaps I'm missing something, but it looks like the changes in this file don't need `tendsto_poly`, so they should be in a different PR?

SpherePacking/ModularForms/EisensteinAsymptotics.lean
b-mehta, line 1: same for this file

SpherePacking/ModularForms/IsCuspForm.lean
b-mehta, line 1: and for this file

SpherePacking/ForMathlib/AtImInfty.lean
cameronfreer: The helper lemmas don't themselves require `tendsto_poly`, but are needed for the applications of `tendsto_poly` that this PR #350 adds. I included them here because I wasn't sure if they were of independent interest besides how they are used here.

But if we'd rather review them separately first (either because we want them anyway, or to keep the review modular), I can move them to a new PR (which the current PR #350 will then depend on in addition to #330). Should I do so?

b-mehta: Yeah I think that's a good idea, in particular because I think @seewoo5 is a better person to review the helper lemmas than I, and conversely for #330.

cameronfreer: Sounds good, thanks. I'll split that off.

I've also rebased both PRs now that the module system has landed, though I'm not sure if I should use modules more extensively in these PRs too.

SpherePacking/ModularForms/FG.lean
```diff
@@ -971,10 +971,10 @@ private theorem D_B_tendsto_zero :
...
       atImInfty (nhds 0) := by
   apply D_tendsto_zero_of_isBoundedAtImInfty (by fun_prop)
-  have h := ((H₂_tendsto_atImInfty.pow 2).const_mul 2).add
-    (((H₂_tendsto_atImInfty.mul H₄_tendsto_atImInfty).const_mul 5).add
-      ((H₄_tendsto_atImInfty.pow 2).const_mul 5))
-  simp only [zero_pow two_ne_zero, one_pow, mul_zero, mul_one, zero_add] at h
+  have := H₂_tendsto_atImInfty
+  have := H₄_tendsto_atImInfty
+  have h : Tendsto (fun z => 2 * H₂ z ^ 2 + 5 * H₂ z * H₄ z + 5 * H₄ z ^ 2)
+      atImInfty (nhds 5) := by tendsto_cont
```
b-mehta, line 977: Here are two extensions you could do to the tactic. Either

a) make an attribute `@[tendsto_cont]` then tagging `H₂_tendsto_atImInfty` means the tactic knows about that limit as a hypothesis (eg how fun_prop does it)
b) let `tendsto_cont` take arguments so `tendsto_cont [H₂_tendsto_atImInfty]` will be able to use that fact as a hypothesis (eg how linarith does it)

I think they could be nicer for applications like this. Seewoo, do you have a preference?

SpherePacking/ModularForms/ThetaDerivIdentities.lean
```diff
@@ -470,31 +469,32 @@ lemma f₄_tendsto_atImInfty : Tendsto f₄ atImInfty (𝓝 0) := by
...
+      atImInfty (𝓝 (1/6 : ℂ)) := by
+    have := H₂_tendsto_atImInfty
+    have := H₄_tendsto_atImInfty
+    tendsto_cont
   simpa [f₄] using h_serre_H₄.add h_scaled
 
 /-- theta_g tends to 0 at infinity.
 theta_g = (2H₂ + H₄)f₂ + (H₂ + 2H₄)f₄.
-Since 2H₂ + H₄ → 1, H₂ + 2H₄ → 2, and f₂, f₄ → 0, we get theta_g → 0. -/
+Using pair-of-pairs: (H₂, H₄) → (0, 1) and (f₂, f₄) → (0, 0), so theta_g → 0. -/
```
b-mehta: imo the earlier docstring was better because it matches what's informally going on (even though secretly the new one is more accurate)

b-mehta (dismissed):

SpherePacking/ModularForms/FG.lean
seewoo5: What I did with `MDifferentiable` in this project is a), see [MDifferentiableFunProp.lean](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/blob/main/SpherePacking/ForMathlib/MDifferentiableFunProp.lean) and [JacobiTheta.lean](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/blob/main/SpherePacking/ModularForms/JacobiTheta.lean).

cameronfreer: How about doing both, like with `simp`? In any case I can start playing around with the two approaches in a new PR.

SpherePacking/ModularForms/ThetaDerivIdentities.lean
cameronfreer: ok, thanks -- reverted in e472f33

SpherePacking/ModularForms/FG.lean
seewoo5: Yes I don't really have preference between two options. In fact, for the MDiff case I'm not sure if that's the idea use of `fun_prop`. I'll merge this one now, and please make new PR if get new approach.

seewoo5 (approved):

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks -- see #378 for an attempt at this

PR #369, opened 2026-03-18
golf: shorten proofs in BoundaryControl.lean

b-mehta:
Compress proofs across BoundaryControl.lean, reducing from 1008 to 866 lines (-142 lines, -14%) while preserving all statements.

Key improvements:
- Triangle inequality proofs use linarith with explicit lemma hints
- Volume bound proofs use calc chains instead of long have sequences
- Boundary strip geometry proof halved via compact coordinate reasoning
- Shell ratio limit proof uses field_simp and direct rw chains
- ENNReal division cancellation proof simplified

### DIFF
--- a/SpherePacking/Basic/PeriodicPacking/BoundaryControl.lean
+++ b/SpherePacking/Basic/PeriodicPacking/BoundaryControl.lean
@@ -519,15 +519,11 @@ theorem exists_periodicSpherePacking_sep_one_density_gt_of_lt_density (hd : 0 < 
   have hcubeShell : cubeShellErr L = shellVol / volCube := by
     simp [cubeShellErr, shellVol, volCube]
   have hvolCube_ne0 : volCube ≠ 0 := by
-    have hvol : volCube = (ENNReal.ofReal L) ^ d := by
-      simpa [volCube] using (PeriodicConstant.volume_coordCube (d := d) (L := L))
-    have hL' : (ENNReal.ofReal L) ≠ 0 := by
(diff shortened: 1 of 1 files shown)


SpherePacking/Basic/PeriodicPacking/BoundaryControl.lean
```diff
@@ -923,13 +645,11 @@ theorem exists_periodicSpherePacking_sep_one_density_gt_of_lt_density (hd : 0 <
...
   have hden :
-      (Real.toNNReal (ZLattice.covolume (cubeLattice (d := d) L hLpos) volume) : ℝ≥0∞) = volCube :=
-      by
-    have : Real.toNNReal (ZLattice.covolume (cubeLattice (d := d) L hLpos) volume) =
-        volCube.toNNReal := by
-      simpa [volCube] using (PeriodicConstantApprox.toNNReal_covolume_cubeLattice
-        (d := d) (L := L) hLpos)
-    simp [this, ENNReal.coe_toNNReal hvolCube_ne_top]
+      (Real.toNNReal (ZLattice.covolume (cubeLattice (d := d) L hLpos) volume) : ℝ≥0∞) = volCube := by
+    rw [show Real.toNNReal (ZLattice.covolume (cubeLattice (d := d) L hLpos) volume) =
```
b-mehta: This is an instance of the banned `rw [show` pattern, please apply the usual fix.

thefundamentaltheor3m (approved):
Thanks for golfing! I do think more golfing might be possible, but one hole at a time I guess 🏌️

PR #374, opened 2026-03-19
run CI on PRs to gauss2

dwrensha:


b-mehta (approved):

b-mehta (commented):
Actually, if it is not a push to main, then we probably don't want to [deploy the website](https://github.com/leanprover-community/docgen-action?tab=readme-ov-file#input-deploy) in docgen-action. Is there an easy way of doing that other than making a different yml?

dwrensha:
docgen-action only deploys on a `push` event:
https://github.com/leanprover-community/docgen-action/blob/7836f3de2f10facc0fed09c1309e6402e4cd5fd0/action.yml#L149

The PR does not change the behavior of `build.yml` on `push` events.

https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/blob/34d457f0818add835c991ccb5f83f898568402d3/.github/workflows/build.yml#L1-L9

arajasek (approved):

b-mehta (approved):

PR #375, opened 2026-03-19
[gauss2] fix linter warnings (except one)

b-mehta:
- Remove deprecated/unused `coe_mk_subtype` from simp in JacobiTheta
- Wrap long comment in Derivative.lean to stay within 100 chars
- Suppress privateModule linter in NormNumI_Scratch (scratch file)


seewoo5 (approved):
Thanks!

PR #378, opened 2026-03-21
feat: `@[tendsto_cont]` attribute + `tendsto_cont [...]` inline arg syntax

cameronfreer:
## Summary

- Add a `@[tendsto_cont]` attribute for registering closed lemmas of the form `Tendsto f l (nhds a)`, so `tendsto_cont` can use them without local `have := ...` boilerplate.
- Add `tendsto_cont [h₁, h₂]` syntax for passing `Tendsto` facts inline.
- Extend candidate collection to use three sources with precedence **inline > local context > attribute registry**. Lower-priority candidates are dropped when a higher-priority source provides the same function, while same-source conflicts still raise the existing ambiguity error.
- Validate `@[tendsto_cont]` eagerly: reject non-`Tendsto` declarations, non-`nhds` targets, parameterized declarations, and non-global (`local`/`scoped`) uses.

## Implementation notes

The attribute registry lives in `TendstoContAttr.lean`, since the tactic code needs the environment extension from a precompiled module.

The cross-bucket shadowing tests use a Sierpinski-style topology on `Bool` via `mkOfNhds`, where `nhds false = ⊤ ≠ pure true = nhds true`. Both `Tendsto` facts are provable (the function always returns `true`), but the neighborhood filters are genuinely distinguishable — so if the tactic accidentally uses the attributed `false` limit when the goal says `nhds true`, `reconcileLimits` fails. The attribute-level ambiguity test uses the indiscrete topology (`⊤`) to register two conflicting limits for the same function without `sorry`.

Suggested by b-mehta in PR #350. `MDifferentiableFunProp.lean` was cited by seewoo5 as prior art.

## Test plan

- [ ] `lake build`
- [ ] `lake build SpherePacking.Tactic.Test.TendstoCont`
- [ ] Positive: `tendsto_cont [h]`, `tendsto_cont [h₁, h₂]`, automatic discovery via `@[tendsto_cont]`, inline shadowing of conflicting local hypotheses, and load-bearing cross-bucket shadowing (attr vs local, attr vs inline) using Sierpinski topology.
- [ ] Negative (`guard_msgs`): wrong-filter and non-`Tendsto` inline args are ignored; conflicting inline facts for the same function error; invalid `@[tendsto_cont]` registrations are rejected (non-`Tendsto`, parameterized, wrong target filter, `local`, `scoped`); attribute-level ambiguity (same function, different limits) errors.


SpherePacking/Tactic/Test/TendstoCont.lean
```diff
@@ -243,3 +243,169 @@ example (hf : Tendsto f (nhds 0) (nhds 1)) :
...
 example (hf : Tendsto f (nhds 0) (nhds 1)) :
     Tendsto (fun x => Real.sin (f x)) (nhds 0) (nhds (Real.sin 1)) := by tendsto_cont
+
+-- ══════════════════════════════════════════════════════════════
+-- Inline argument syntax: tendsto_cont [h₁, h₂]
+-- ══════════════════════════════════════════════════════════════
+
+-- Single inline argument
+example (h : Tendsto f atTop (nhds 3)) :
+    Tendsto (fun z => f z + 1) atTop (nhds 4) := by tendsto_cont [h]
```
b-mehta: Doesn't this example also work with just `tendsto_cont`, because the tactic already uses local hypotheses? 
(In fact, in this case, perhaps we should even warn against this, since `h` is redundant)

SpherePacking/Tactic/Test/TendstoCont.lean
```diff
@@ -243,3 +243,169 @@ example (hf : Tendsto f (nhds 0) (nhds 1)) :
...
+-- ══════════════════════════════════════════════════════════════
+
+-- Register a closed Tendsto theorem via attribute
+@[tendsto_cont]
+private theorem testAttrLemma :
+    Tendsto (fun z : ℝ => z) (nhds 0) (nhds 0) :=
+  tendsto_id
+
+-- Attribute-registered theorem used automatically (no local hypothesis needed)
+example : Tendsto (fun z : ℝ => z + 1) (nhds 0) (nhds 1) := by tendsto_cont
```
b-mehta: I'm also not sure what this example shows, since `fun_prop` could deduce this one, and (I believe) it works without lines 268-271. I think a better test-case here would be to `axiom`atise a function `f : \R -> \R`, and `[tendsto_cont]` that `Tendsto f (nhds 0) (nhds 0)`, then check `Tendsto (fun z : ℝ => f z + 1) (nhds 0) (nhds 1) := by tendsto_cont` works.

b-mehta (commented):
I haven't fully read the code yet, but the positive examples are currently not so convincing. I hope this is because of poor examples rather than because of poor code, but I'll review the code once the examples are doing what they ought to do!

SpherePacking/Tactic/Test/TendstoCont.lean
cameronfreer: Yes, thanks for catching that -- it now has `private def`/`private theorem` examples where the `Tendsto` facts aren't local hypotheses. There's now a `#guard_msgs` test showing bare `tendsto_cont` fails (no candidates), followed by `tendsto_cont [inlineFn_tendsto]` succeeding in 2c88647. 

I'll add the warning for the previous circumstance in a later commit.

SpherePacking/Tactic/Test/TendstoCont.lean
cameronfreer: Yes, this was also provable without the attribute. There's now a version with concrete definitions (`private def attrFn` + `private theorem attrFn_tendsto`) in 2c88647 where a `#guard_msgs` test illustrates that `tendsto_cont` fails before attribute `[tendsto_cont] attrFn_tendsto`, then succeeds after.

I'll also add an axiom example in a later commit.

SpherePacking/Tactic/Test/TendstoCont.lean
cameronfreer: I've added warnings in 77bea00 and b64926f.

SpherePacking/Tactic/Test/TendstoCont.lean
cameronfreer: There's an axiom example in [a422b5b](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/378/commits/a422b5b88a276a7f7477532274b3337eb93a1e65).

SpherePacking/Tactic/TendstoCont.lean
```diff
@@ -141,20 +142,86 @@ private meta partial def findAtomsAux (e : Expr) (bvar : FVarId)
...
+  -- Bucket 1: inline args (highest priority)
+  let mut inlineCands : Array Atom := #[]
+  for hyp in extraHyps do
+    let ty ← inferType hyp >>= instantiateMVars
+    match ← matchTendstoNhds? ty with
+    | some (codTy, f, l, a) =>
+      if ← withNewMCtxDepth (isDefEq l goalFilter) then
+        inlineCands := inlineCands.push
+          { fn := f, limit := a, hyp := hyp, codTy := codTy }
+    | none => continue
```
b-mehta: Can we make each bucket its own function for readability and logic separation?

b-mehta (dismissed):
Looks good to me, with one small style point.

SpherePacking/Tactic/TendstoCont.lean
cameronfreer: I recently did something similar in #384 (in response to the [zulip discussion of `fun_prop` improvements](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/computing.20derivatives.20with.20fun_prop/near/581632950)).  I've now backported some of that in 4f39b8f.

seewoo5 (approved):
I tested this with #307, and this works for me. But I may wait until all @b-mehta's concerns are resolved.
```lean
lemma denominator_tendsto_at_infty :
    Tendsto (fun s ↦ (H₄.resToImagAxis s).re ^ 3 *
      (2 * (H₄.resToImagAxis s).re ^ 2 + 5 * (H₂.resToImagAxis s).re * (H₄.resToImagAxis s).re
        + 5 * (H₂.resToImagAxis s).re ^ 2)) atTop (nhds 2) := by
  tendsto_cont [H₂_resToImagAxis_tendsto_zero, H₄_resToImagAxis_tendsto_one]
```

seewoo5:
@b-mehta Do you have any further comments on this PR? Also @cameronfreer it might be great if this can be upstreamed to mathlib (in some form).

b-mehta:
Oops, sorry I missed this! I think we should merge this, since it's at least useful here, and then any further improvements can happen later. 

Separately, I think @lecopivo was thinking about an extensible thing somewhat like this for mathlib.

thefundamentaltheor3m:
Merging modulo CI after merging main! Thanks all!

PR #385, opened 2026-03-31
[gauss2] fix linter warnings

dwrensha:
Fixes all linter warnings except the one about the too-long import line in the toplevel `SpherePacking.lean`.


b-mehta (approved):
Thanks!

PR #413, opened 2026-04-29
refactor(E8): tidy up basic E8 lattice file

seewoo5:
## Summary
- Remove the `section hack` workaround (`Matrix.myDet`, `E8Matrix_myDet_eq_one`, `ZSpan.volume_fundamentalDomain'`); `ZSpan.volume_fundamentalDomain` is usable directly.
- Rename placeholder helpers `test''`, `same_domain`, `E8_Basis_volume` to descriptive names and mark internal ones `private`.
- Small golfing pass: collapse multi-line `rw; rfl`, merge `simp_rw` with `simp only`, drop redundant tactics, pack split signatures (net: +21/-45 lines).

## Test plan
- [x] `lake build SpherePacking.Basic.E8` succeeds
- [x] No new diagnostics in `SpherePacking/Basic/E8.lean`
- [x] Downstream callers of `E8Packing_density` / `E8_ℤBasis_ofZLatticeBasis_volume` (the renamed `E8_Basis_volume`) still build

### DIFF
--- a/SpherePacking/Basic/E8.lean
+++ b/SpherePacking/Basic/E8.lean
@@ -104,12 +104,10 @@ lemma Submodule.coe_evenLattice (R : Type*) (n : ℕ) [Ring R] [CharZero R] :
     exact hm
   · simpa [funext_iff]
 
-lemma Submodule.mem_evenLattice {R : Type*} [Ring R] [CharZero R] (n : ℕ)
-    {v : Fin n → R} :
+lemma Submodule.mem_evenLattice {R : Type*} [Ring R] [CharZero R] (n : ℕ) {v : Fin n → R} :
     v ∈ Submodule.evenLattice R n ↔
       (∀ i, ∃ n : ℤ, (n : R) = v i) ∧ ∑ i, v i ≡ 0 [PMOD 2] := by
-  rw [← SetLike.mem_coe, Submodule.coe_evenLattice]
-  rfl
+  rw [← SetLike.mem_coe, Submodule.coe_evenLattice]; rfl
 
 -- TODO (BM): this shouldn't be noncomputable... check with @zwarich
 noncomputable def Submodule.E8 (R : Type*) [Field R] [NeZero (2 : R)] :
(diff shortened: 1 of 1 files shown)


SpherePacking/Basic/E8.lean
```diff
@@ -650,34 +642,18 @@ lemma E8_ℤBasis_apply_norm : ∀ i : Fin 8, ‖E8_ℤBasis i‖ ≤ 2 := by
   intro i
   simpa [coe_E8_ℤBasis_apply, E8Basis_apply] using E8Basis_apply_norm i
 
-section hack
```
b-mehta, line 653: oh yeah, mathlib's reducibility on Matrix.det changed, so the hack isn't needed any more! nice catch

SpherePacking/Basic/E8.lean
```diff
@@ -686,10 +662,10 @@ lemma same_domain :
   simp [E8_ℤBasis, E8Basis_apply]
 
 open MeasureTheory ZSpan in
-lemma E8_Basis_volume :
+lemma E8_ℤBasis_ofZLatticeBasis_volume :
     volume (fundamentalDomain (E8_ℤBasis.ofZLatticeBasis ℝ E8Lattice)) = 1 := by
-  rw [← test'']
-  erw [same_domain]
+  rw [← volume_preimage_toLp_symm]
+  erw [preimage_fundamentalDomain_E8_ℤBasis]
```
b-mehta: can we remove the erw now?

b-mehta (dismissed):

SpherePacking/Basic/E8.lean
seewoo5: It seems not - changing `erw` to `rw` gives
```
Tactic `rewrite` failed: Did not find an occurrence of the pattern
  ⇑(WithLp.linearEquiv 2 ℝ ((i : Fin 8) → (fun x ↦ ℝ) i)).symm ⁻¹'
    fundamentalDomain (Basis.ofZLatticeBasis ℝ E8Lattice E8_ℤBasis)
in the target expression
  volume
      (⇑(WithLp.equiv 2 ((i : Fin 8) → (fun x ↦ ℝ) i)).symm ⁻¹'
        fundamentalDomain (Basis.ofZLatticeBasis ℝ E8Lattice E8_ℤBasis)) =
    1
```
But it works if we change the previous lemma as
```lean
private lemma preimage_fundamentalDomain_E8_ℤBasis :
    (WithLp.equiv 2 _).symm ⁻¹' fundamentalDomain (E8_ℤBasis.ofZLatticeBasis ℝ E8Lattice) =
      fundamentalDomain (E8Basis ℝ) := by
  change (WithLp.linearEquiv 2 ℝ (Fin 8 → ℝ)).symm ⁻¹' _ = _
  rw [← LinearEquiv.image_eq_preimage_symm, ZSpan.map_fundamentalDomain]
  congr! 1
  ext i : 1
  simp [E8_ℤBasis, E8Basis_apply]
```
Would this be better?

b-mehta: Yeah! (it's not ideal, but it's definitely better)

seewoo5: Updated! Just for curiosity, what should we do if we also want to remove the extra `change` line?

b-mehta: I think the answer would be to be more consistent with using WithLp.equiv or WithLp.linearEquiv, and maybe changing some bits of mathlib to be more flexible in what they accept. (But after your question I looked deeper, and it's trickier than I guessed at first)

b-mehta (approved):

PR #416, opened 2026-05-19
Prepare defs for mathlib

thefundamentaltheor3m:
This PR aims to prepare `SpherePacking.Basic.SpherePacking.lean` to be PRed to `mathlib`.

Let's use this thread to discuss any changes we want to implement before PRing to `mathlib`, such as generalising definitions. My understanding from our last packathon is that we want to stick to Euclidean Space because sphere packings can be poorly behaved in unusual metric spaces. I can include that as an implementation note if that's something we think would be useful.

See also [my Mathlib fork](https://github.com/thefundamentaltheor3m/mathlib4/commit/d4961fffcf3d1cc920e89a51b8007957c2889a0c#diff-72eb2a8fc0cf3b99a0853c0a6d6d8c0eca6f5f828c51220094c92603bab8bb5a).

### DIFF
--- a/SpherePacking/Basic/SpherePacking.lean
+++ b/SpherePacking/Basic/SpherePacking.lean
@@ -17,8 +17,6 @@ public import SpherePacking.ForMathlib.VolumeOfBalls
 
 @[expose] public section
 
-open BigOperators MeasureTheory Metric
-
 /-!
 # Density of Sphere Packings
 
@@ -29,17 +27,21 @@ the sphere packing centers.
 We also define the *density* of the configuration.
 -/
 
+open BigOperators MeasureTheory Metric Pointwise Filter Module
 open scoped ENNReal
-open BigOperators Pointwise Filter Module
 
 section Definitions
 
+/-- A sphere packing in dimension `d` is a set of points in `ℝ^d` (the *centers*) such that any two
+distinct points are at distance at least `separation` apart. -/
 structure SpherePacking (d : ℕ) where
   centers : Set (EuclideanSpace ℝ (Fin d))
   separation : ℝ
   separation_pos : 0 < separation := by positivity
   centers_dist : Pairwise (separation ≤ dist · · : centers → centers → Prop)
 
+/-- A periodic sphere packing in dimension `d` is a sphere packing whose centers are invariant
+under translation by a `ℤ`-lattice in `ℝ^d`. -/
 structure PeriodicSpherePacking (d : ℕ) extends SpherePacking d where
   lattice : Submodule ℤ (EuclideanSpace ℝ (Fin d))
   lattice_action : ∀ ⦃x y⦄, x ∈ lattice → y ∈ centers → x + y ∈ centers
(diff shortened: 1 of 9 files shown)


thefundamentaltheor3m:
Note that we can remove [`VolumeOfBalls.lean`](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/blob/b6567c28712260b0c0765e127a691ea10d1938fd/SpherePacking/ForMathlib/VolumeOfBalls.lean).

seewoo5:
Shouldn't it be not too hard to work with general metric space? I thought that is useful to other projects since people work on sphere packing in hyperbolic space or discrete setting (coding theory, hamming distance, etc). Lattice packing is (in my opinion) only useful in the Euclidean setting, or other spaces may have different definition of "lattices".

b-mehta:
Yeah I agree general metric space is the way to go

thefundamentaltheor3m:
Makes sense. I think we said during one of the packathons that we weren't sure how sensible that'd be because densities might not be as well behaved in other metric spaces. But let me see what I can adapt to the general setting: maybe not everything, but maybe not nothing either. It may also be possible to define Lattice Packings slightly more generally (eg. any (metric) space in which lattices exist). Let me give that a try.

thefundamentaltheor3m:
Comments on generality are being addressed in #417.

SpherePacking/Basic/SpherePacking.lean
```diff
@@ -6,80 +6,126 @@ Authors: Sidharth Hariharan, Gareth Ma
...
-the sphere packing centers.
-
-We also define the *density* of the configuration.
+# Sphere Packings
+
+A *sphere packing* in `ℝᵈ` is a set of *centers* `X ⊆ ℝᵈ` with a positive `separation` such
+that distinct centers are at distance at least `separation`. Placing an open ball of radius
+`separation / 2` around each center yields the packing's (disjoint) balls. In this file, we define
+sphere packings and periodic sphere packings in `ℝᵈ` and the notions of sphere packing density and
+the sphere packing constant and prove basic facts about these definitions.
```
seewoo5, line 22: Docstring may needs to be updated for general metric spaces (and possibly below descriptions, too)

SpherePacking/Basic/SpherePacking.lean
```diff
@@ -6,80 +6,126 @@ Authors: Sidharth Hariharan, Gareth Ma
...
-  lattice : Submodule ℤ (EuclideanSpace ℝ (Fin d))
+/-- A periodic sphere packing in a normed  a sphere packing whose centers are invariant
+under translation by a `ℤ`-lattice in `ℝ^d`. -/
+structure PeriodicSpherePacking (K E : Type*) [NormedField K] [NormedAddCommGroup E]
+    [NormedSpace K E] extends SpherePacking E where
+  lattice : Submodule ℤ E
   lattice_action : ∀ ⦃x y⦄, x ∈ lattice → y ∈ centers → x + y ∈ centers
   lattice_discrete : DiscreteTopology lattice := by infer_instance
-  lattice_isZLattice : IsZLattice ℝ lattice := by infer_instance
+  lattice_isZLattice : IsZLattice K lattice := by infer_instance
```
seewoo5, line 90: I think this is lattice packing, not periodic packing, where latter allows a finite union of translation of a lattice (such as Figure 5 [in this note](https://arxiv.org/pdf/1611.01685)).

Related to the last meeting's discussion, I think that the best definition for periodic packing for general metric space is when it is (freely) act by certain group $G$, then we take a discrete subgroup $\Gamma$ and consider finite union of orbits of points and packing centered at these configurations. For example, Euclidean space $\mathbb{R}^d$ is act by itself by translation, and taking $\Gamma = \mathbb{Z}^d$ gives the usual periodic packing. But other interesting case would be hyperbolic packing, e.g. when $X = \mathbb{H}^2$ and $G = \mathrm{PSL}_2(\mathbb{R}) \supset \mathrm{PSL}_2(\mathbb{Z}) = \Gamma$. Although I believe implementing this general definition might be too much for now, this current definition for $K$-vector space might be only useful for Euclidean space and may not give new use case.

seewoo5: For the first part I misread the previous line - I thought it is `x ∈ centers → y ∈ centers → x + y ∈ centers` instead of `x ∈ lattice → y ∈ centers → x + y ∈ centers`, so let me take it back. But maybe we want to prove that this definition is equivalent to another definition of periodic packing, as finite union of translation of a lattice? (Also, would it be helpful to have `LatticePacking`?)

PR #423, opened 2026-06-19
refactor(RealDecay): consolidate decay lemmas into one general helper

cameronfreer:
## Summary

Consolidate the exponential-decay `Tendsto` helpers in `RealDecay.lean` into one `n`-parametrized lemma:

- **Add** `tendsto_const_mul_pow_mul_exp_neg_atTop (C a : ℝ) (n : ℕ) (ha : 0 < a)` — `C · tⁿ · exp(-a·t) → 0`.
- **Remove** the three single-power lemmas it subsumes (each is the `C = 1` case): `tendsto_exp_neg_atTop` (n=0), `tendsto_mul_exp_neg_atTop` (n=1), `tendsto_sq_mul_exp_neg_atTop` (n=2).

The removed lemmas had no remaining uses — the vertical/top-edge decay proofs in [#296](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/296) route through the general lemma.

## How this helps #296

#296 (ContourEndpoints) proves `tendsto_verticalBound_atTop` / `tendsto_topEdgeBound_atTop` term by term, each of the form `C · tⁿ · exp(-a·t)` (`n = 0, 1, 2`). With the general lemma every term is a one-line application instead of a five-line `const_mul / simp / convert / ring` block (`n = 2` directly, `n = 0, 1` via `simpa`).

## Notes

- **Must merge before #296.** #296 carries the identical RealDecay change so it stays buildable in the meantime; that change drops out once this merges.

## Test plan
- [x] `lake build` passes
- [x] No sorries; no new axioms

### DIFF
--- a/SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
+++ b/SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
@@ -27,7 +27,7 @@ These lemmas are designed to be reusable for contour integral analysis where:
 ## Main results
 
 ### Asymptotic behavior
-- `tendsto_const_mul_pow_mul_exp_neg_atTop`: C · tⁿ · exp(-a·t) → 0 as t → ∞ for a > 0
+- `tendsto_const_mul_rpow_mul_exp_neg_atTop`: C · t^n · exp(-a·t) → 0 as t → ∞ for a > 0
 
 ### Integrability
 - `integrableOn_exp_mul_Ici`: exp(c*t) is integrable on [1,∞) for c < 0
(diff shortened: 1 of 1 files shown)


SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
```diff
@@ -55,20 +53,11 @@ lemma integrableOn_exp_mul_Ici (c : ℝ) (hc : c < 0) :
...
-lemma tendsto_sq_mul_exp_neg_atTop (a : ℝ) (ha : 0 < a) :
-    Tendsto (fun t => t^2 * exp (-a * t)) atTop (nhds 0) := by
-  simpa [rpow_two] using tendsto_rpow_mul_exp_neg_mul_atTop_nhds_zero 2 a ha
-
-/-- exp(-a*t) → 0 as t → ∞ for a > 0. -/
-lemma tendsto_exp_neg_atTop (a : ℝ) (ha : 0 < a) :
-    Tendsto (fun t => exp (-a * t)) atTop (nhds 0) := by
-  simpa [rpow_zero] using tendsto_rpow_mul_exp_neg_mul_atTop_nhds_zero 0 a ha
+/-- `C · tⁿ · exp(-a·t) → 0` as `t → ∞` for `a > 0` (covers the `n = 0, 1, 2` decay terms). -/
+lemma tendsto_const_mul_pow_mul_exp_neg_atTop (C a : ℝ) (n : ℕ) (ha : 0 < a) :
```
seewoo5: We can allow `n` to be arbitrary real, right? I guess most of applications in this project may only need natural number version, but I don't think it hurts to give general statements (unless you need to make nontrivial downstream change).

SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
```diff
@@ -55,20 +53,11 @@ lemma integrableOn_exp_mul_Ici (c : ℝ) (hc : c < 0) :
...
+    Tendsto (fun t => C * t ^ n * exp (-a * t)) atTop (nhds 0) := by
```
seewoo5: Use `\mapsto` instead of `=>` (I recently found that Claude usually do `=>` unless we explicitly tell it)

seewoo5 (dismissed):
Minor comments!

SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
cameronfreer: Yes, that worked, thanks. The proof here simplified a bit; two applications in #296 now need to use `Real.rpow_two` but that was the only change.

SpherePacking/MagicFunction/a/Integrability/RealDecay.lean
cameronfreer: Ah, thanks. Now fixed, and tracked in [lean4-skills#133](https://github.com/cameronfreer/lean4-skills/issues/133).

seewoo5 (approved):

PR #331, opened 2026-02-03
Prove FG_inequality_1 and FG_inequality_2

seewoo5:
Prove two modular form inequalities.
- `FmodG_AntitoneOn` is changed into `FmodG_strictAntiOn`
- Depends on #307 

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #428, opened 2026-06-23
feat(ResToImagAxis): inv/div + quotient-real-part API; use in FG

cameronfreer:
## Summary

`ResToImagAxis.Real F` ("`F` is real on the positive imaginary axis") had a full `@[fun_prop]` closure suite for `neg/add/sub/mul/smul/pow` but **no `inv`/`div` and no real-part projection** for products/quotients. Because of that gap, several proofs hand-rolled the imaginary-axis quotient/product real-part step (`FG.lean`'s private `div_re_of_im_eq_zero` + an 8-line block; the manual `mul_re` blocks inside `Pos.mul` / `EventuallyPos.mul`). This PR adds the missing reusable API and uses it.

## Changes

**`ResToImagAxis.lean`** — new lemmas beside the existing closure suite:
- `@[fun_prop] Real.inv : Real F → Real F⁻¹` and `@[fun_prop] Real.div : Real F → Real G → Real (F/G)` (the latter now derived from `inv` + `mul`). No `≠ 0` hypothesis is needed — `(w⁻¹).im = -w.im/normSq` is `0` when `w.im = 0` (even at `w = 0`, since `0⁻¹ = 0`).
- `Real.re_mul_eq : ((F*G).resToImagAxis t).re = (F.resToImagAxis t).re * (G.resToImagAxis t).re` (both factors real).
- `Real.re_div_eq : ((F/G).resToImagAxis t).re = (F.resToImagAxis t).re / (G.resToImagAxis t).re` — needs **only** the denominator's realness.
- `Real.re_div_mul_eq` — the `F/(G*H)` version, a two-rewrite corollary of the two above.
- `Real.div_eq_real_div : F(it)/G(it) = ↑((F(it)).re / (G(it)).re)` for real `F`, `G`.
- Relocated the `(a/b).re = a.re/b.re` helper `div_re_of_im_eq_zero` here from `FG.lean`.

Projection lemmas (`re_mul_eq`/`re_div_eq`/`re_div_mul_eq`/`div_eq_real_div`) are intentionally **not** `@[simp]` — they're explicit rewrite tools.

**In-file reuse (demonstrates the API):** `Pos.mul` and `EventuallyPos.mul` each drop their manual `mul_re`/realness block — the product-positivity step is now `rw [Real.re_mul_eq …]; exact mul_pos …`.

**`FG.lean`:**
- Collapsed the quotient block in `L₁₀_div_FG_tendsto` (8 lines manually rebuilding `(F·G)`'s realness) to one `re_div_mul_eq F_imag_axis_real G_imag_axis_real t`; the `filter_upwards` no longer needs `eventually_gt_atTop 0` (the identity holds for all `t`).
- Removed the now-relocated private `div_re_of_im_eq_zero`.
- Removed unneeded `have hGh := G_holo` in `L₁₀_imag_axis_real`.
- **Closed the `FmodG_eq_FmodGReal` sorry** via the new `div_eq_real_div`, and dropped its now-unused (ht : 0 < t) hypothesis — it has no callers (main or any monotonicity branch) and is vacuously true for all t:
  ```lean
  unfold FmodGReal FReal GReal
  exact (ResToImagAxis.Real.div_eq_real_div F_imag_axis_real G_imag_axis_real t).symm
  ```
  (For contrast, the *existing* API already supported it, but in more steps:)
  ```lean
  unfold FmodGReal FReal GReal
  rw [ResToImagAxis.Real.eq_real_part F_imag_axis_real t,
    ResToImagAxis.Real.eq_real_part G_imag_axis_real t]
  simp [Complex.ofReal_div]
  ```
  The lemma is relocated from the early definition block to just after `F`/`G_imag_axis_real` (the proof needs those realness facts; it has no other uses in the file).

## Where this is reusable downstream

This rounds out a small, coherent "real part of an operation on the imaginary axis" API. Concretely:
- `@[fun_prop] Real.inv`/`Real.div` — these complete the `@[fun_prop]` realness closure suite (alongside `neg`/`add`/`sub`/`mul`/`smul`/`pow`). No call site exercises them yet — this PR's callers invoke the projection lemmas (`re_div_eq`, `div_eq_real_div`, …) directly, since they need an equality of real parts rather than just ResToImagAxis.Real (F/G). So they're forward-looking, but they do fire — `fun_prop` discharges each of these from the base realness facts:
```
example : ResToImagAxis.Real (F / G) := by
  have hF := F_imag_axis_real; have hG := G_imag_axis_real; fun_prop
example : ResToImagAxis.Real G⁻¹ := by
  have hG := G_imag_axis_real; fun_prop
example : ResToImagAxis.Real (L₁₀ / (F * G)) := by
  have hL := L₁₀_imag_axis_real; have hF := F_imag_axis_real; have hG := G_imag_axis_real; fun_prop
```
- **`re_mul_eq`/`re_div_eq`/`re_div_mul_eq`** — wherever a real part of a product/quotient on the imaginary axis is computed by hand. Already applied here to `Pos.mul`, `EventuallyPos.mul`, and the `L₁₀` block.
- **`div_eq_real_div`** — the bridge between a complex quotient `F(it)/G(it)` and the real quotient `FReal/GReal`. Beyond `FmodG_eq_FmodGReal` (closed here), this is the natural tool for the `FmodGReal` monotonicity / `F`-over-`G` work (PR #331, branch `monotonicity-of-F-div-G`), e.g. `FReal = FmodGReal * GReal`-style identities.

Caveat: #331's functional-equation realification mostly rewrites individual values with `eq_real_part` and then does coercion/ring work, so `re_mul_eq` helps it only where expressions are kept in `(F * G).resToImagAxis` product form — a real but partial win there.

## Test plan
- [x] `lake build` passes — one **fewer** sorry (`FmodG_eq_FmodGReal`); the unrelated WIP sorries in `FG`/`PolyFourierCoeffBound` are untouched.

## Note re PR #331
#331 (FG inequalities) adds only S-action lemmas to `ResToImagAxis.lean` (no overlap with this quotient/product API) and leaves the `L₁₀` block / `FmodG_eq_FmodGReal` as-is, so these rebase cleanly — verify against the live #331 head at merge time.

### DIFF
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -561,6 +561,7 @@ theorem G_imag_axis_pos : ResToImagAxis.Pos G := by unfold G; fun_prop (disch :=
 `G(it)` is real for all `t > 0`.
 Blueprint: G = H₂³ (2H₂² + 5H₂H₄ + 5H₄²), product of real functions.
 -/
+@[fun_prop]
 theorem G_imag_axis_real : ResToImagAxis.Real G := G_imag_axis_pos.1
 
 /--
@@ -576,6 +577,7 @@ theorem F_imag_axis_pos : ResToImagAxis.Pos F := by
 `F(it)` is real for all `t > 0`.
 Blueprint: Follows from E₂, E₄, E₆ having real values on the imaginary axis.
 -/
+@[fun_prop]
 theorem F_imag_axis_real : ResToImagAxis.Real F := F_imag_axis_pos.1
 
 /-- `FmodGReal t = F(it)/G(it)` on the imaginary axis. (Placed here, after `F`/`G_imag_axis_real`,
@@ -1024,8 +1026,6 @@ theorem D_G_div_G_tendsto :
 /-- `L₁,₀(it)` is real for all `t > 0`. -/
 theorem L₁₀_imag_axis_real : ResToImagAxis.Real L₁₀ := by
   unfold L₁₀
-  have hF := F_imag_axis_real
-  have hG := G_imag_axis_real
   fun_prop
 
 /-- `lim_{t→∞} L₁,₀(it)/(F(it)G(it)) = 1/2`. -/
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -1051,7 +1051,7 @@ theorem L₁₀_div_FG_tendsto :
     (tendsto_resToImagAxis_of_tendsto_atImInfty h_L_over_FG)
   simp only [show (1 / 2 : ℂ).re = (1 / 2 : ℝ) by norm_num] at h_re
   refine h_re.congr' ?_
-  filter_upwards [Filter.eventually_gt_atTop 0] with t _ht_pos
+  filter_upwards with t
   simp only [Function.comp_apply]
   exact ResToImagAxis.Real.re_div_mul_eq F_imag_axis_real G_imag_axis_real t
 
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -561,7 +561,6 @@ theorem G_imag_axis_pos : ResToImagAxis.Pos G := by unfold G; fun_prop (disch :=
 `G(it)` is real for all `t > 0`.
 Blueprint: G = H₂³ (2H₂² + 5H₂H₄ + 5H₄²), product of real functions.
 -/
-@[fun_prop]
 theorem G_imag_axis_real : ResToImagAxis.Real G := G_imag_axis_pos.1
 
 /--
@@ -577,7 +576,6 @@ theorem F_imag_axis_pos : ResToImagAxis.Pos F := by
 `F(it)` is real for all `t > 0`.
 Blueprint: Follows from E₂, E₄, E₆ having real values on the imaginary axis.
 -/
-@[fun_prop]
 theorem F_imag_axis_real : ResToImagAxis.Real F := F_imag_axis_pos.1
 
 /-- `FmodGReal t = F(it)/G(it)` on the imaginary axis. (Placed here, after `F`/`G_imag_axis_real`,
@@ -877,8 +875,9 @@ theorem D_G_div_G_tendsto :
 
 /-- `L₁,₀(it)` is real for all `t > 0`. -/
 theorem L₁₀_imag_axis_real : ResToImagAxis.Real L₁₀ := by
-  unfold L₁₀
-  fun_prop
+  have hF := F_imag_axis_real
+  have hG := G_imag_axis_real
+  fun_prop [L₁₀]
 
 /-- `lim_{t→∞} L₁,₀(it)/(F(it)G(it)) = 1/2`. -/
 theorem L₁₀_div_FG_tendsto :
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -580,7 +580,7 @@ theorem F_imag_axis_real : ResToImagAxis.Real F := F_imag_axis_pos.1
 
 /-- `FmodGReal t = F(it)/G(it)` on the imaginary axis. (Placed here, after `F`/`G_imag_axis_real`,
 since the proof needs both realness facts.) -/
-theorem FmodG_eq_FmodGReal {t : ℝ} (ht : 0 < t) :
+theorem FmodG_eq_FmodGReal {t : ℝ} :
     FmodGReal t = (F.resToImagAxis t) / (G.resToImagAxis t) := by
   unfold FmodGReal FReal GReal
   exact (ResToImagAxis.Real.div_eq_real_div F_imag_axis_real G_imag_axis_real t).symm
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -73,9 +73,6 @@ theorem F_eq_FReal {t : ℝ} (ht : 0 < t) : F.resToImagAxis t = FReal t := by so
 
 theorem G_eq_GReal {t : ℝ} (ht : 0 < t) : G.resToImagAxis t = GReal t := by sorry
 
-theorem FmodG_eq_FmodGReal {t : ℝ} (ht : 0 < t) :
-    FmodGReal t = (F.resToImagAxis t) / (G.resToImagAxis t) := by sorry
-
 /--
 `F = 9 * (D E₄)²` by Ramanujan's formula.
 From `ramanujan_E₄`: `D E₄ = (1/3) * (E₂ * E₄ - E₆)`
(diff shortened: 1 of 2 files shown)


SpherePacking/ModularForms/FG.lean
```diff
@@ -586,6 +578,14 @@ Blueprint: Follows from E₂, E₄, E₆ having real values on the imaginary axi
 -/
 theorem F_imag_axis_real : ResToImagAxis.Real F := F_imag_axis_pos.1
 
+/-- `FmodGReal t = F(it)/G(it)` on the imaginary axis. (Placed here, after `F`/`G_imag_axis_real`,
+since the proof needs both realness facts.) -/
+theorem FmodG_eq_FmodGReal {t : ℝ} (ht : 0 < t) :
+    FmodGReal t = (F.resToImagAxis t) / (G.resToImagAxis t) := by
+  have _ := ht
```
seewoo5: I guess this is unnecessary?

SpherePacking/ModularForms/ResToImagAxis.lean
```diff
@@ -145,6 +145,54 @@ theorem ResToImagAxis.Real.pow {F : ℍ → ℂ} (hF : ResToImagAxis.Real F) (n
...
+    ResToImagAxis.Real F⁻¹ := by
+  intro t ht
+  have hFreal := hF t ht
+  simp only [Function.resToImagAxis, ResToImagAxis, ht, ↓reduceDIte] at hFreal
+  simp [ResToImagAxis, ht, Pi.inv_apply, Complex.inv_im, hFreal]
+
+@[fun_prop]
+theorem ResToImagAxis.Real.div {F G : ℍ → ℂ} (hF : ResToImagAxis.Real F)
+    (hG : ResToImagAxis.Real G) : ResToImagAxis.Real (F / G) := by
+  simpa [div_eq_mul_inv] using hF.mul hG.inv
```
seewoo5, line 159: Are these `@[fun_prop]` registrations actually used? It seems that you directly call these theorems rather applying `fun_prop` in other files. I'm fine with either way - keep tags and try to replace with `fun_prop`, or remove tags.

SpherePacking/ModularForms/FG.lean
cameronfreer: I think originally I had thought it was needed to avoid a warning because `ht` is not used, but it works fine without it. (Now dropped in 4824d32.)

But separately, perhaps we should change this theorem by removing `(ht : 0 < t)` entirely -- I hesitated to do that here because it changes the API, but is that (along with any downstream consequences) worth doing in a separate PR?

SpherePacking/ModularForms/ResToImagAxis.lean
cameronfreer: They're not used here directly (and if we keep them, I should highlight that in the PR description). In particular, the quotient projection proof calls `div_eq_real_div` directly because it needs an equality, not just `ResToImagAxis.Real (F / G)`.

But I think they could be helpful elsewhere; e.g., they enable these to work:

```
example : ResToImagAxis.Real (F / G) := by
  have hF := F_imag_axis_real
  have hG := G_imag_axis_real
  fun_prop

example : ResToImagAxis.Real G⁻¹ := by
  have hG := G_imag_axis_real
  fun_prop

example : ResToImagAxis.Real (L₁₀ / (F * G)) := by
  have hL := L₁₀_imag_axis_real
  have hF := F_imag_axis_real
  have hG := G_imag_axis_real
  fun_prop
```

I don't see specific places we could use it to simplify existing code. Should I remove the tags for now and we can revisit this later if/when the tags would be used?

SpherePacking/ModularForms/FG.lean
cameronfreer: Actually, I don't think there are downstream consumers of this, so probably it's best to just remove the hypothesis `(ht : 0 < t)`.

SpherePacking/ModularForms/ResToImagAxis.lean
seewoo5: I think it is better to keep these.

SpherePacking/ModularForms/FG.lean
```diff
@@ -1027,7 +1026,6 @@ theorem L₁₀_imag_axis_real : ResToImagAxis.Real L₁₀ := by
   unfold L₁₀
   have hF := F_imag_axis_real
   have hG := G_imag_axis_real
-  have hGh := G_holo
   fun_prop
```
seewoo5: Now we can do `fun_prop [F_imag_axis_real, G_imag_axis_real]`?

seewoo5 (dismissed):

SpherePacking/ModularForms/FG.lean
cameronfreer: That didn't work, but tagging the two remaining lemmas with `@[fun_prop]` does -- see [6811beb](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/428/commits/6811beb81f5d8db6852094452d0c1647286d211e). I was a bit confused about it; below is what Claude explained in light of the various experiments.

Do we want the tags added? It has some other benefits, like allowing the examples `Real (F/G)` and `Real G⁻¹` to close with just `fun_prop`. Or should I revert 6811beb?



> Here's the mechanism, based on how fun_prop is architected (and consistent with exactly what your experiments showed):
> 
> fun_prop decomposes to "leaves", then discharges each leaf two specific ways
> 
> fun_prop proves a goal like ResToImagAxis.Real (D F * G - F * D G) by structural recursion on the function expression, applying @[fun_prop] composition rules (Real.sub, Real.mul, the D-preserves-realness rule, …) to peel off the operations. It bottoms out at leaves — atomic, single declared functions like F and G (after η-expansion, fun a ↦ F a). To finish, it must discharge each leaf goal Real F, Real G.
> 
> For a leaf, it has exactly two sources of facts:
> 
> 1. The @[fun_prop] index — a discrimination tree of function theorems, keyed by the leaf's head constant. When you tag F_imag_axis_real, it's registered here under F; the leaf lookup for Real F finds it. (This is the same index that holds E₄'s realness, which is why DE₄_imag_axis_real := by fun_prop works.)
> 2. The local context — as a fallback, the leaf discharger checks local hypotheses for a matching fact. That's why have hF := F_imag_axis_real works: hF : Real F sits in context and the fallback picks it up.
> 
> The error you saw — No theorems found for F in order to prove ResToImagAxis.Real fun a ↦ F a — is precisely source 1 coming up empty: the leaf reached F, the index had no function theorem keyed under F, and there was no local hypothesis either.
> 
> Why the bracket and disch don't reach the leaves
> 
> - fun_prop [F_imag_axis_real, G_imag_axis_real] — the bracket feeds extra theorems into the composition/transition machinery for that call, not into the leaf-discharge function-theorem index. So F_imag_axis_real is available as a "rewrite/transition rule," but the leaf lookup for Real F never consults it. (A bare ground fact like Real F isn't a composition rule, so it just sits unused.) Net effect: the index is still empty under F → "No theorems found for F."
> - disch := … exact F_imag_axis_real — disch is the discharger for side conditions: the hypotheses of the composition rules (things like a b ≠ 0 premise on a division rule, or a measurability side-goal). It is not invoked on the function-property leaves themselves. The recursion handles Real F via sources 1/2 above and fails before any disch would run, so your exact F_imag_axis_real discharger never fires.
> 
> So the only two things that actually inform the leaf are
> 
> - registering the fact in the index (@[fun_prop]), or
> - putting it in the local context (have).
> 
> Both work; the bracket and disch target different stages of the algorithm (composition rules and side conditions, respectively), so they can't substitute for either. Tagging is the "global" version of the have trick, which is why it both cleans up L₁₀_imag_axis_real and lets Real (F/G)/Real G⁻¹ close bare (now the F/G leaves resolve from the index, and Real.div/Real.inv are already-registered composition rules).

seewoo5: IMO I want to keep `fun_prop` for the "general" theorems but not for specific functions, except for basic building blocks (Eisenstein series and theta function). In particular, the specific F and G here won't be the case. So I think it would be better to revert it.
But I still want to understand what is happening here...

cameronfreer: Makes sense. I reverted the tags but also saved one line because `L₁₀` does work as a parameter to `fun_prop` rather than unfolding it earlier. See 0a1a02a.

seewoo5 (approved):
Thanks!

PR #429, opened 2026-06-23
refactor(FG): extract log-derivative-at-infinity keystone

cameronfreer:
## Summary

Three "log-derivative at infinity" limits in `FG.lean` — `D_F_div_F_tendsto`, `D_G_div_G_tendsto`, `D_H₂_div_H₂_tendsto` — each had a bespoke ~30–40-line proof rebuilding the same idea: a modular form `F` behaves like `exp(a·z) · (nonzero limit factor)` near i∞, so the log-derivative `D F / F → a/(2πi)`. With `Derivative.lean`'s `D_tendsto_zero_of_isBoundedAtImInfty` now available, the vanishing bounded-factor term is a one-call fact, so the whole pattern packages into a single keystone.

## Changes

- **Add `logderiv_tendsto_of_div_exp_tendsto`** (public): if `F z / exp(a·z) → C ≠ 0` at i∞, then `D F / F → a/(2πi)`. The exponential contributes `a/(2πi)`; the bounded limit factor's log-derivative vanishes.
- **Golf the three theorems** to a keystone call + a uniform `a/(2πi)` reduction, consuming the existing `F_vanishing_order` / `G_vanishing_order` / `H₂_div_exp_tendsto` facts directly. Statements are unchanged (`D_F_div_F_tendsto` / `D_G_div_G_tendsto` are still consumed by `L₁₀_div_FG_tendsto`).
- **Move `D_cexp_div`** above `F_vanishing_order` so the keystone can use it.
- **Delete the now-orphaned scaffolding**: `D_diff_div_q_tendsto`, `D_exp_pi_div_exp_pi`, `D_B_tendsto_zero`, `D_H₂_tendsto_zero`, `D_H₄_tendsto_zero`, `H₂_eventually_ne_zero`.
- **Second commit — golf**: the keystone's two continuity-of-limit steps (the quotient `D g / g → 0` and adding the constant exponential contribution) are discharged by `tendsto_cont`; the genuinely-pointwise algebraic part (`logderiv_mul_eq` + eventual equality) is unchanged.

Net **−140 lines** in `FG.lean`. (`D_diff_qexp` is kept though now unused — a public q-expansion-of-`D` lemma; happy to drop it if preferred.)

## Reusability

The keystone is **public**, so any file importing `FG.lean` can reuse the same proof pattern once they have a normalized exponential asymptotic.
- **#331** gets the concrete fourth site: `D_Θ₂_div_Θ₂_tendsto`. On rebase, its current bounded-factor proof should collapse to the same call, using its `Θ₂_div_exp_tendsto : Θ₂ z /  exp(πiz/4) → 2`:  `logderiv_tendsto_of_div_exp_tendsto Θ₂_MDifferentiable (by norm_num) Θ₂_div_exp_tendsto`, followed by `(πI/4)/(2πI) = 1/8`.
- The lemma also covers the `a = 0` case: if `F → C ≠ 0`, then `D F / F → 0`. So future `D H₃ / H₃` or `D H₄ / H₄` limits would be one-call consequences of `H₃_tendsto_atImInfty` / `H₄_tendsto_atImInfty` plus the existing MDiff facts, if those goals ever appear.
- Not direct targets: `EisensteinAsymptotics.lean` uses `D_tendsto_zero_of_isBoundedAtImInfty`, but its current goals are Serre-derivative/asymptotic limits rather than log-derivative quotients; the `MagicFunction` files do not have this exponential-main-term log-derivative shape.

Lower-level files such as `JacobiTheta/JacobiIdentity.lean` cannot call this helper while it lives in `FG.lean`, because `FG.lean` already imports them. If theta-product/Delta log-derivative limits are later needed there, that is the reason to move the helper, together with `D_cexp_div` / `logderiv_mul_eq`, to `Derivative.lean`.

## Test plan
- [x] `lake build` passes; `L₁₀_div_FG_tendsto` and `JacobiIdentity.lean` still compile; no new sorries.

## Note re overlap
This and the open `restoimagaxis-real-div` PR both touch `FG.lean` (different regions — the asymptotics block here vs the L₁₀/FmodG block there). Rebase on whichever of that PR / main lands first.

### DIFF
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -762,16 +762,11 @@ lemma logderiv_tendsto_of_div_exp_tendsto {F : ℍ → ℂ} (hF : MDiff F) {a C 
   have hq_ne : ∀ w : ℍ, q w ≠ 0 := fun w ↦ Complex.exp_ne_zero _
   have hq_md : MDiff q := by
     intro τ
-    have h_diff : DifferentiableAt ℂ (fun t : ℂ ↦ cexp (a * t)) (τ : ℂ) :=
-      (differentiableAt_id.const_mul a).cexp
-    simpa [hq, Function.comp] using
-      DifferentiableAt_MDifferentiableAt (G := fun t : ℂ ↦ cexp (a * t)) (z := τ) h_diff
+    simpa [hq, Function.comp] using DifferentiableAt_MDifferentiableAt
+      (G := fun t : ℂ ↦ cexp (a * t)) (z := τ) ((differentiableAt_id.const_mul a).cexp)
   have hg_md : MDiff g := MDifferentiable_div hF hq_md hq_ne
-  have hg_lim : Filter.Tendsto g atImInfty (nhds C) := hlim
-  have hDg_tendsto : Filter.Tendsto (D g) atImInfty (nhds 0) :=
-    D_tendsto_zero_of_isBoundedAtImInfty hg_md (hg_lim.isBigO_one ℝ)
   have hDg_div_g : Filter.Tendsto (fun z ↦ D g z / g z) atImInfty (nhds 0) := by
-    tendsto_cont
+    simpa using (D_tendsto_zero_of_isBoundedAtImInfty hg_md (hlim.isBigO_one ℝ)).div hlim hC
   have hF_eq : F = q * g := by ext w; simp only [hg, Pi.mul_apply, mul_div_cancel₀ _ (hq_ne w)]
   have hDq_div_q : ∀ z : ℍ, D q z / q z = a / (2 * π * I) :=
     fun z ↦ by rw [hq]; exact D_cexp_div a z
@@ -779,8 +774,7 @@ lemma logderiv_tendsto_of_div_exp_tendsto {F : ℍ → ℂ} (hF : MDiff F) {a C 
     filter_upwards [hlim.eventually_ne hC] with z hz
     rw [hF_eq, Pi.mul_apply, logderiv_mul_eq q g hq_md hg_md z (hq_ne z) hz, hDq_div_q z]
   have hsum : Filter.Tendsto (fun z ↦ a / (2 * π * I) + D g z / g z) atImInfty
-      (nhds (a / (2 * π * I))) := by
-    tendsto_cont
+      (nhds (a / (2 * π * I))) := by tendsto_cont
   exact hsum.congr' (by filter_upwards [h_logderiv] with z hz; exact hz.symm)
 
 /-- The vanishing order of F at infinity is 2.
--- a/SpherePacking/ModularForms/FG.lean
+++ b/SpherePacking/ModularForms/FG.lean
@@ -740,8 +740,44 @@ private theorem H₂_div_exp_tendsto :
     simp only [H₂, div_pow, ← Complex.exp_nat_mul]; congr 2; ring
   simp_rw [h_eq]; convert Θ₂_div_exp_tendsto.pow 4; norm_num
 
-private lemma H₂_eventually_ne_zero : ∀ᶠ z : ℍ in atImInfty, H₂ z ≠ 0 :=
-  eventually_ne_zero_of_tendsto_div (by norm_num : (16 : ℂ) ≠ 0) H₂_div_exp_tendsto
+/-- D(exp(c*z))/exp(c*z) = c/(2πi) for any coefficient c. -/
+theorem D_cexp_div (c : ℂ) (z : ℍ) :
+    D (fun w => cexp (c * w)) z / cexp (c * z) = c / (2 * π * I) := by
+  simp only [D]
+  have h_deriv : deriv ((fun w : ℍ => cexp (c * w)) ∘ ⇑ofComplex) (z : ℂ) =
+      c * cexp (c * z) :=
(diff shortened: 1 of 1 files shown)


SpherePacking/ModularForms/FG.lean
```diff
@@ -799,89 +839,12 @@ theorem D_diff_qexp (z : ℍ) :
...
-    have h_eq : ∀ z : ℍ, D f z / f z = (D f z / cexp (2 * π * Complex.I * z)) /
-        (f z / cexp (2 * π * Complex.I * z)) := fun z => by field_simp [Complex.exp_ne_zero]
-    simp_rw [h_eq, show (1 : ℂ) = 720 / 720 from by norm_num]
-    exact hDf_div_q.div hf_div_q h_720_ne
-  have h_F_ne := eventually_ne_zero_of_tendsto_div
-    (by norm_num : (720^2 : ℂ) ≠ 0) F_vanishing_order
-  simpa using (hDf_div_f.const_mul (2 : ℂ)).congr' (by
-    filter_upwards [h_F_ne] with z hFz; exact (hDF_div_eq z hFz).symm)
+  have h := logderiv_tendsto_of_div_exp_tendsto F_holo (by norm_num) F_vanishing_order
+  rwa [show (2 * π * I * 2) / (2 * π * I) = 2 from by field_simp] at h
```
seewoo5:
```suggestion
  rwa [show (2 * π * I * 2) / (2 * π * I) = 2 by field_simp] at h
```
I saw this a lot from Claude, too.

cameronfreer: Thanks -- fixed, and noted in [lean4-skills#133](https://github.com/cameronfreer/lean4-skills/issues/133#issuecomment-4779322649).
I also fixed a few more `=>` vs. `↦`.

SpherePacking/ModularForms/FG.lean
```diff
@@ -740,8 +740,48 @@ private theorem H₂_div_exp_tendsto :
...
+    Filter.Tendsto (fun z : ℍ ↦ D F z / F z) atImInfty (nhds (a / (2 * π * I))) := by
+  set q : ℍ → ℂ := fun w ↦ cexp (a * w) with hq
+  set g : ℍ → ℂ := fun w ↦ F w / q w with hg
+  have hq_ne : ∀ w : ℍ, q w ≠ 0 := fun w ↦ Complex.exp_ne_zero _
+  have hq_md : MDiff q := by
+    intro τ
+    have h_diff : DifferentiableAt ℂ (fun t : ℂ ↦ cexp (a * t)) (τ : ℂ) :=
+      (differentiableAt_id.const_mul a).cexp
+    simpa [hq, Function.comp] using
+      DifferentiableAt_MDifferentiableAt (G := fun t : ℂ ↦ cexp (a * t)) (z := τ) h_diff
```
seewoo5: Probably we can inline `h_diff`?

SpherePacking/ModularForms/FG.lean
```diff
@@ -740,8 +740,48 @@ private theorem H₂_div_exp_tendsto :
...
+  have hg_md : MDiff g := MDifferentiable_div hF hq_md hq_ne
+  have hg_lim : Filter.Tendsto g atImInfty (nhds C) := hlim
```
seewoo5: Redefining same thing here

SpherePacking/ModularForms/FG.lean
```diff
@@ -740,8 +740,48 @@ private theorem H₂_div_exp_tendsto :
...
+    tendsto_cont
+  have hF_eq : F = q * g := by ext w; simp only [hg, Pi.mul_apply, mul_div_cancel₀ _ (hq_ne w)]
+  have hDq_div_q : ∀ z : ℍ, D q z / q z = a / (2 * π * I) :=
+    fun z ↦ by rw [hq]; exact D_cexp_div a z
+  have h_logderiv : ∀ᶠ z : ℍ in atImInfty, D F z / F z = a / (2 * π * I) + D g z / g z := by
+    filter_upwards [hlim.eventually_ne hC] with z hz
+    rw [hF_eq, Pi.mul_apply, logderiv_mul_eq q g hq_md hg_md z (hq_ne z) hz, hDq_div_q z]
+  have hsum : Filter.Tendsto (fun z ↦ a / (2 * π * I) + D g z / g z) atImInfty
+      (nhds (a / (2 * π * I))) := by
+    tendsto_cont
```
seewoo5:
```suggestion
  have hsum : Filter.Tendsto (fun z ↦ a / (2 * π * I) + D g z / g z) atImInfty
      (nhds (a / (2 * π * I))) := by tendsto_cont
```

SpherePacking/ModularForms/FG.lean
cameronfreer: Fixed -- cleaner now, but one less opportunity to use `tendsto_cont`.

SpherePacking/ModularForms/FG.lean
cameronfreer: Yes, thanks

SpherePacking/ModularForms/FG.lean
cameronfreer: Thanks, fixed

cameronfreer:
Thanks, I also golfed a bit more.

seewoo5 (approved):
Thanks!

PR #437, opened 2026-07-14
blueprint: Refine sections on Fourier analysis, Cohn–Elkies and the Fourier eigenfunctions

thefundamentaltheor3m:
# Blueprint sync with the M4R thesis

This PR updates the blueprint (mainly §7 on the Fourier eigenfunctions, and to a lesser extent §4 on Fourier analysis and §5 on the Cohn–Elkies bound) to follow the detailed treatment in Hariharan's MSci thesis, including its Mathlib references, with one deliberate departure: the Schwartzness bridge across dimensions is stated via a **smooth cutoff**, following the architecture of (the description of) draft PR #433 rather than the thesis's smooth-on-`[0,∞)` composition theorem.

## §4 — Facts from Fourier analysis

* `lemma:Gaussian-Fourier` now has a proof, via Mathlib's `fourier_gaussian_innerProductSpace` with `b = -πiz` (no branch issues since `d/2 = 4` is an integer).
* New subsection on **radial functions and Fourier eigenfunctions**: the involution `𝓕² f = f(-·)` (`fourier_involution`, proved), `𝓕² = id` on even/radial Schwartz functions (`radial_inversion`, proved), and the decomposition `f = f₊ + f₋` into `±1`-eigenfunctions that drives the construction of `a` and `b`.
* New subsection **"Radial Schwartz functions from one-dimensional data"**, replacing the old `thm:smooth-fast-decay-schwartz` (whose global hypotheses the one-dimensional integrals do not satisfy — they grow exponentially as `r → -∞`, and `I₆`/`J₆` are not even convergent integrals far to the left):
  - composition with `‖·‖²` sends `𝓢(ℝ, ℂ)` to `𝓢(ℝᵈ, ℂ)` (`schwartzMap_multidimensional_of_schwartzMap_real`, proved, via Doll's `SchwartzMap.compCLM`);
  - a smooth transition function `Θ` (`Real.smoothTransition`), `0` on `(-∞, -1]` and `1` on `[0, ∞)`;
  - **`thm:cutoff-schwartz`**: if `f` is smooth on `(θ, ∞)` (`θ < 0`) with all derivatives decaying rapidly on `[0, ∞)`, then `Θ_θ · f` is Schwartz — with a full proof of *why* (locality of smoothness on the cover `(-∞, θ/2) ∪ (θ, ∞)`; Leibniz rule plus the fact that all derivatives of the cutoff vanish on `[0, ∞)` and are supported in the compact transition interval);
  - the corollary that `x ↦ f(‖x‖²)` is Schwartz, since the cutoff is invisible after composing with `‖·‖² ≥ 0`.

## §5 — Cohn–Elkies

* New subsection **"The shape of an optimal function"** (following Cohn's ICM discussion, as in the thesis): equality analysis of the proof forces vanishing at non-zero lattice points, double zeroes, and `g(0) = ĝ(0)`; radiality and the eigenfunction decomposition reduce the problem to constructing `±1`-eigenfunctions.
* The rescaling argument after `thm:g` is now spelled out (and corrected: `f(x) = g(√2 x)`, not `g(x/√2)`), with the `16 · Vol(B₈(0,½)) = π⁴/384` arithmetic via Mathlib's volume-of-balls formula.

## §7 — Fourier eigenfunctions with double zeroes

* **Formalisation links added throughout**: `φ₀/φ₂'/φ₄'`, `φ₀_periodic`/`φ₀_S_transform` (proved), the integral estimates `Iⱼ'_bounding` (I₁/I₃/I₅ fully proved), `h`, `ψI/ψT/ψS`, the ψ-identities `ψI_eq`/`ψT_eq`/`ψS_eq'` and slash relations (all proved), the permutation lemmas `perm_I₁_I₂`, `perm_I₅` (+ `J`-analogues), their reverse directions via radial inversion (proved), `eig_a`/`eig_b`, and the unbounded rectangular Cauchy–Goursat lemmas from `ForMathlib/CauchyGoursat`.
* New `lemma:poly-growth-mul` (closure of polynomial-growth Fourier expansions under products) with the explicit `n₀` computations used to derive the `φ`- and `ψ`-bounds from `lemma:mod-div-disc-bound`, as in the thesis.
* New `lem:Ij-smooth`, recording smoothness via the Leibniz integral rule and the convergence thresholds that set the cutoff parameters: `I₆` is smooth only on `(-2, ∞)`, `J₆` only on `(-1, ∞)` — hence cutoffs with `θ = -2` for `a` and `θ = -1` for `b`.
* `prop:a-schwartz`/`prop:b-schwartz` re-proved via the cutoff bridge; the eigenfunction propositions restructured into permutation lemmas matching the Lean architecture (Fubini + Gaussian Fourier + Möbius substitution + rectangular Cauchy–Goursat).
* `lemma:psi-slash-relations` and `lemma:psi-sum` (`ψT + ψS = ψI`) extracted as standalone lemmas; `lemma:psiI-psiT-psiS-fourier` and the previously proof-less `J`-bounds given proofs.
* **Corrections**: sign errors in `lemma:bound-J1-J3-J5` (`e^{πr/s} → e^{-πr/s}`) and `lemma:bound-J2-J4-J6` (`e^{π(r+1)} → e^{-π(r+1)}`); `prop:a0` corrected from `a(0) = -i/8640` to `a(0) = -8640i/π` (matching `MagicFunction.a.SpecialValues.a_zero` and the thesis); dangling reference `eqn:a-definition` fixed; special-value proofs actually compute the limits.
* `thm:g1` linked to the formalised `g`, `g_zero`, `fourier_g_zero`, with the final computation `g(0) = ĝ(0) = 1` spelled out.

## Verification

* `leanblueprint web` (plasTeX) parses cleanly with all subsections inlined — no new warnings.
* `latexmk` (xelatex) compiles `print.pdf` with **no errors, no undefined or multiply-defined references**.
* Every `\lean{…}` name added was checked to exist on `main` or in Mathlib `v4.30.0`; `\leanok` markers were only added where the corresponding declaration/proof is actually sorry-free on `main` (e.g. the perm-lemma *statements* are marked, their sorried forward proofs are not).
* No Lean files are touched, so `lake build` is unaffected.

https://claude.ai/code/session_01SDUxbBrwnHuT2DrS9Axai3

---
_Generated by [Claude Code](https://claude.ai/code/session_01SDUxbBrwnHuT2DrS9Axai3)_


thefundamentaltheor3m:
Independent audit of this PR (commit 0a5b205):

- **b-side sign chain**: with `def:b-definition`, prop:b-double-zeros forces `b(r) = −4i sin²(πr²/2)∫₀^∞ψ_I(it)e^{−πr²t}dt` (verified numerically to 10 digits from the contour integrals). Fixed prop:b-another-integral (`4i → −4i`) and thm:g1: the magic function is `g = (πi/8640)a − (i/(240π))b` (simple zero at √2, slope −√2/60); the `+` combination is `ĝ`. Lean `g` and `J₆'` (`+2` vs `−2`, contradicting Lean's own `perm_J₅`/`b_zero`) flagged for matching fixes.
- eqn:ineqBnew direction corrected to `F − 18π⁻²G < 0` (matches its proof and `FG_inequality_2`); restored missing `t²` in the equivalence lemma.
- Cutoff rescaling fixed to `Θ(2x/(−θ))` (transition on `[θ/2, 0]`, as in PR #433).
- ψ_I near-0 bound: `O(t²e^{−π/t})`; lem:integral-bound generalised to `α > 0` (covers `e^{−πs}`); sign typos in a-side proofs.
- PDF, plasTeX web and all 324 `\uses` edges verified clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

blueprint/src/subsections/fourier-analysis.tex
```diff
@@ -12,21 +12,27 @@
 
 The following computational result will be of use later on.
 
-\begin{lemma}\label{lemma:Gaussian-Fourier}\uses{def:Fourier-Transform}
+\begin{lemma}\label{lemma:Gaussian-Fourier}\uses{def:Fourier-Transform}\lean{fourier_gaussian_innerProductSpace}
+  Let $z \in \h$ and let $G_z : \R^8 \to \C$ denote the Gaussian $G_z(x) := e^{\pi i \|x\|^2 z}$. Then, for all $y \in \R^8$,
```
seewoo5: nitpick: G_z is only introduced/used in this lemma and if it is intended it might be natural to write L18 as $G_{-1/z}$, but it didn't, so I think original blueprint is fine

blueprint/src/subsections/fourier-analysis.tex
```diff
@@ -36,6 +42,42 @@ \subsection{On Schwartz Functions}
...
+\begin{lemma}\label{lemma:radial-fourier-inversion}\uses{def:radial-function, lemma:fourier-involution}\lean{MagicFunction.a.Fourier.radial_inversion}\leanok
+  If $f : \R^d \to \C$ is an even Schwartz function --- in particular, if $f$ is radial --- then $\mathcal{F}(\mathcal{F}(f)) = f$.
+\end{lemma}
+\begin{proof}\leanok
+  Immediate from Lemma~\ref{lemma:fourier-involution}, since $f(-x) = f(x)$.
+\end{proof}
+
+Lemma~\ref{lemma:radial-fourier-inversion} says that $\mathcal{F}$ is an involution on radial Schwartz functions, so such functions decompose into eigenfunctions of $\mathcal{F}$ with eigenvalues $\pm 1$.
+
+\begin{lemma}\label{lemma:eigenfunction-decomposition}\uses{lemma:radial-fourier-inversion}
```
seewoo5, line 72: To be strict, we don't need this theorem in our project since we are going to explicitly construct $f$. This is more about the strategy to construct $f$.

However, one can consider the following general version: if you have an additive group with involution $(M, inv)$ where 2 is invertible in $M$ and $inv(inv(x)) = x$ (and $inv$ is homomorphism from $M$ to itself). Then any element can be written as "symmetric" and "anti-symmetric" elements: $x = (x + inv(x)) /2 + (x - inv(x)) / 2$.

blueprint/src/subsections/modform-ineq.tex
```diff
@@ -51,7 +51,7 @@ \section{Proof of Theorem \ref{thm:g}}\label{sec: g}
 Inequality \eqref{eqn:ineqA} and \eqref{eqn:ineqB} are equivalent to
 \begin{align}
     F(it) + \frac{18}{\pi^2} G(it) > 0 \label{eqn:ineqAnew} \\
-    F(it) - \frac{18}{\pi^2} G(it) > 0 \label{eqn:ineqBnew}
+    F(it) - \frac{18}{\pi^2} G(it) < 0 \label{eqn:ineqBnew}
```
seewoo5, line 54: 👍

blueprint/src/subsections/modform-ineq.tex
```diff
@@ -220,36 +220,49 @@ \section{Proof of Theorem \ref{thm:g}}\label{sec: g}
...
+% the magic function is (πi/8640)a - (i/(240π))b: for r > √2 one has
+% a(r) = 4i sin²(πr²/2)∫₀^∞ t²φ₀(i/t)e^{-πr²t}dt and
+% b(r) = -4i sin²(πr²/2)∫₀^∞ ψ_I(it)e^{-πr²t}dt (Propositions \ref{prop:a-double-zeros} and
+% \ref{prop:b-double-zeros}), so only the minus combination produces the integrand A(t) < 0.
+% The Lean declaration `g` currently uses the plus sign and must be corrected accordingly.
+\begin{theorem}\label{thm:g1}\lean{g, g_zero, fourier_g_zero}
 \uses{prop:a-fourier, prop:b-fourier, prop:a-double-zeros, prop:b-double-zeros, prop:ineqA, prop:ineqB, prop:a0, prop:b0}
 The function
-$$g(x):=\frac{\pi\,i}{8640}a(x)+\frac{i}{240\pi}\,b(x)$$
+$$g(x):=\frac{\pi\,i}{8640}a(x)-\frac{i}{240\pi}\,b(x)$$
```
seewoo5, line 245: original sign is correct?

blueprint/src/subsections/fourier-analysis.tex
thefundamentaltheor3m: I guess LLMs like new (and unnecessary) definitions even in natural language... I guess in this case it's more "notation" than a "definition" but it's still unnecessary

blueprint/src/subsections/fourier-analysis.tex
thefundamentaltheor3m: This will be addressed in #444. I will make another PR outlining what I plan to do mathematically (for the blueprint).

blueprint/src/subsections/modform-ineq.tex
thefundamentaltheor3m: Yes. It was a typo in my thesis. It is now fixed!

blueprint/src/subsections/cohn-elkies.tex
viazovska, line 1: This definition:

A $C^\infty$~function $f:\R^d\to\C$ is called a \emph{Schwartz function} if it decays to zero as $\|x\|\to\infty$ faster then any inverse power of $\|x\|$, and the same holds for all partial derivatives of $f$, ie, if for all $k, n \in \N$, there exists a constant $C \in \R$ such that for all $x \in \R^d$, $\norm{x}^k \cdot \norm{f^{(n)}(x)} \leq C$, where $f^{(n)}$ denotes the $n$-th derivative of $f$ considered along with the appropriate operator norm. The set of all Schwartz functions from $\R^d$ to $\C$ is called the \emph{Schwartz space}. It is an $\R$-vector space.

looks a bit cryptic to me. I wold prefer:

A $C^\infty$~function $f:\R^d\to\C$ is called a \emph{Schwartz function} if it decays to zero as $\|x\|\to\infty$ faster then any inverse power of $\|x\|$, and the same holds for all partial derivatives of $f$, ie, if for all $k \in \N$ and  $n \in \N^d$, there exists a constant $C \in \R$ such that for all $x \in \R^d$, $\norm{x}^k \cdot \abs{f^{(n)}(x)} \leq C$, where $f^{(n)}$ denotes the $n$-th partial derivative of $f$. The set of all Schwartz functions from $\R^d$ to $\C$ is called the \emph{Schwartz space}. It is an $\R$-vector space.

Over all, both versions are acceptable.

blueprint/src/subsections/modform-ineq.tex
seewoo5: I think the sign is still wrong (unwanted fix from + to -)..?

thefundamentaltheor3m: The `−` is forced by `def:b-definition` as it stands (with $J_6 = -2\int_i^{i\infty}\psi_S$):

1. `prop:b-double-zeros` (unchanged from the original blueprint) gives $b(r) = -4\sin^2(\pi r^2/2)\int_0^{i\infty}\psi_I(z)e^{\pi i r^2 z}\,dz$ for $r>\sqrt2$; its proof needs exactly $\psi_T-\psi_I=-\psi_S$, matching $J_6$'s sign.
2. Parametrising $z = it$ gives $dz = i\,dt$, hence the prefactor $-4i$. Unlike the $a$-side there is no $z^2 = -t^2$ factor to absorb the $i^2$ — that's why $a$ gets $+4i$ but $b$ gets $-4i$.
3. Numerical check at $r^2 = 3$: $\sum_j J_j = -0.200307084824\,i$, and $-4i\sin^2(\pi r^2/2)\int_0^\infty \psi_I(it)e^{-3\pi t}dt$ agrees to $10^{-27}$, while the $+4i$ version has the opposite sign.

Then matching $\frac{\pi}{2160}A(t)$ with $A<0$ (so that $g\le 0$ beyond $\sqrt2$) forces $g = \frac{\pi i}{8640}a - \frac{i}{240\pi}b$; the $+$ combination is exactly $\widehat{g}$. The $+$ convention would instead require flipping the sign of $J_6$ (equivalently of $b$), which would then contradict `prop:b-double-zeros` and `perm_J₅ : 𝓕(J₅) = −J₆` in the Lean code.

---
_Generated with [Claude Code](https://claude.ai/code)_

---
_Generated by [Claude Code](https://claude.ai/code)_

blueprint/src/subsections/fourier-analysis.tex
viazovska, line 90: Then, there exists $k \in \N$ sufficiently large such that

can be replaced with 

any real $k>d$

thefundamentaltheor3m:
TODO: Update to get this in line with the latest smooth transition functions done with Bhavik #444, #460 etc

PR #440, opened 2026-07-18
refactor: remove dead files and theorems

seewoo5:
Mostly existed for E2, which all now on mathlib.


SpherePacking/ModularForms/limunder_lems.lean
```diff
@@ -1,60 +0,0 @@
...
-
-@[expose] public section
-
-open TopologicalSpace Set
-  Metric Filter Function Complex
-
-open scoped Interval Real NNReal ENNReal Topology BigOperators Nat
-
-
-lemma limUnder_add {α : Type*} [Preorder α] [(atTop : Filter α).NeBot] (f g : α → ℂ)
```
seewoo5, line 20: @CBirkbeck I wonder if these `limUnder` theorems are upstreamed or planned to be upstreamed. I'm pretty sure that the other theorems are auxiliary theorems for E2, but these seem quite general anyway. Otherwise we can probably merge this.

CBirkbeck: Apparently we don't want more limunder results in mathlib (at least that's what happened when I tried to PR these things).

CBirkbeck:
LGTM!

SpherePacking/ModularForms/limunder_lems.lean
b-mehta: wait, who said this? do you have a link?

CBirkbeck (dismissed):

SpherePacking/ModularForms/limunder_lems.lean
CBirkbeck: https://leanprover.zulipchat.com/#narrow/channel/144837-PR-reviews/topic/.2329115.3A.20develop.20some.20more.20limUnder.20API/with/538645886

CBirkbeck: Maybe I am exaggerating a bit, but when I tried to PR some API we ended up doing a lot of other stuff instead

thefundamentaltheor3m (approved):

Issue #80, opened 2025-06-13
Proving that the periodic sphere packing constant equals the sphere packing constant in all dimensions


thefundamentaltheor3m:

YaelDillies:
claim

alreadydone:
I looked at the proof in Cohn--Elkies Appendix A and found it unsatisfactorily relies on results in another paper [Gr] (19 pages in German). I was able to work out a proof that doesn't rely on such notions as "uniformly dense packing":

Given an arbitrary packing $P$ of upper density $\Delta$ and an arbitrary $\epsilon>0$, we want to produce a periodic packing with density at least $\Delta-\epsilon$. Given a set $S$ (of nonzero volume), we shall call the quantity $\frac{|S\cap P|}{|S|}$ the density of $P$ in $S$, where $|S|$ denotes the volume of $S$.

Choose $\ell>0$ such that $\frac{(\ell-4r)^d}{\ell^d}>1-\frac\epsilon3$ (where $r$ is the radius of the balls in $P$ and $d$ is the dimension). Then choose $R_0>0$ such that for all $R>R_0$, we have $\frac{(R+\sqrt{d}\ell)^d-(R-\sqrt{d}\ell)^d}{R^d}<\frac\epsilon3$. By definition of upper density, there exists $R>R_0$ such that the density of $P$ in $B(0,R)$ is at least $\Delta-\frac\epsilon3$.

[...]

PR #441, opened 2026-07-18
blueprint: add Junyan's proof of periodic constant equals constant

thefundamentaltheor3m:
Taken from #80.

Co-Authored-By: @alreadydone


seewoo5 (approved):
Thanks! Just curious; have you already tried to (auto)formalize this proof? I'm curious about validity of proof - I skimmed through it and looks correct to me, but it would be great to see this formally before we cleanup the proof.

thefundamentaltheor3m:
Yes, I have a (bad but correct) "vibe-proof" of this. I think we can try cleaning it up with Fable or something, but I also think Gonzalo wanted to give it a try.

PR #442, opened 2026-07-18
Bump Lean to v4.32.0

seewoo5:
Update mathlib to v4.32.0 (81a5d257) and the toolchain to leanprover/lean4:v4.32.0, and repair the resulting breakage:

- NoAtoms → NullSingletonClass: mathlib renamed the measure class and left `NoAtoms` only as an `alias`, which is no longer usable as an instance binder (VolumeOfBalls, PeriodicPacking).
- ForMathlib/Encard: drop the local `ENat.add_eq_top`, now upstreamed into mathlib; keep `add_ne_top` (its `by simp` still uses mathlib's @[simp] `add_eq_top`).
- Tactic/NormNumI: `deriveBool` now returns a dependent `BoolResult`; consume it with `match (dependent := true)` so the proof type refines per branch.
- deriv_denom_zpow: mathlib added `UpperHalfPlane.deriv_denom_zpow`, clashing with the project's root-level lemma under `open UpperHalfPlane`; qualify the call sites with `_root_.` (Derivative, SerreDerivativeSlash).


SpherePacking/ModularForms/Eisenstein.lean
```diff
@@ -374,112 +374,7 @@ theorem E₂_imag_axis_real : ResToImagAxis.Real E₂ := by
...
-      rw [← Complex.exp_nat_mul]; ring_nf
-    simp [this]
-  calc ‖1 - 24 * ∑' n : ℕ+, ↑n * cexp (2 * π * Complex.I * ↑n * ↑z) /
-          (1 - cexp (2 * π * Complex.I * ↑n * ↑z))‖
-      = ‖1 - 24 * S‖ := by rw [hS_eq]
-    _ ≤ 1 + 24 * ‖S‖ := by
-        calc _ ≤ ‖(1 : ℂ)‖ + ‖24 * S‖ := norm_sub_le _ _
-          _ = _ := by simp
-    _ ≤ 1 + 24 * (r₀ / (1 - r₀) ^ 3) := by
-        gcongr; exact norm_tsum_logDeriv_expo_le_of_norm_le hq_bound hr₀_lt_one
```
thefundamentaltheor3m, line 489: Presumably these lines can be deleted because the code has been upstreamed?

SpherePacking/ModularForms/EisensteinAsymptotics.lean
```diff
@@ -61,52 +61,43 @@ lemma modular_form_tendsto_atImInfty {k : ℤ} (f : ModularForm (Gamma 1) k) :
     (periodic_comp_ofComplex f hΓ)]
   simpa using (tendsto_zero_of_exp_decay hc hO).add_const (valueAtInfty f.toFun)
 
+/-- `E₂ ∘ ofComplex` is 1-periodic (off `ℍ` both sides are the same junk value). -/
+private lemma E₂_periodic_comp_ofComplex : Function.Periodic (E₂ ∘ ofComplex) 1 := by
```
thefundamentaltheor3m, line 65: Should be `E₂_comp_ofComplex_periodic` or something. But also why is a new lemma being added on a bump PR?

SpherePacking/Tactic/NormNumI.lean
```diff
@@ -170,10 +170,10 @@ such that `norm_num` successfully recognises both the real and imaginary parts o
   let ⟨az, bz, pfz⟩ ← NormNumI.parse z
   let ⟨aw, bw, pfw⟩ ← NormNumI.parse w
   let ⟨ba, ra⟩ ← deriveBool q($az = $aw)
-  match ba with
+  match (dependent := true) ba with
```
thefundamentaltheor3m, line 179: I don't actually know what this means... need to take a look

thefundamentaltheor3m (commented):
Mostly ok, just trying to understand what the big deleted and added lemmas are (I also want to see what `dependent := true` means)

SpherePacking/ModularForms/Eisenstein.lean
seewoo5: Half true and half false. The statement is upstreamed (see [here](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/ModularForms/EisensteinSeries/E2/Summable.html#EisensteinSeries.isBoundedAtImInfty_E2)). However, the proof is slightly different; especially, it does not depend on the Lambert series expansion $E_2 = 1 + 24 \sum_{n \ge 1} q^n / (1 - q^n)$. The latter expansion itself is upstreamable though.

SpherePacking/ModularForms/EisensteinAsymptotics.lean
seewoo5: It seems following [this](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/ModularForms/QExpansion.html#SlashInvariantFormClass.periodic_comp_ofComplex) theorem's name, and possibly added to give better proof of `E₂_sub_one_isBigO_exp` below, and also `E₂_tendsto_one_atImInfty`. (I may upstream it later)

SpherePacking/Tactic/NormNumI.lean
seewoo5: I also want to understand this... FYI, here's Fable's take:

---

Background: what this code is. evalComplexEq is a norm_num extension — a metaprogram that runs while the tactic executes. It doesn't manipulate complex numbers; it manipulates expressions (Expr, Lean syntax trees) representing them, and it must construct an Expr that is a valid proof term. The Q(...)/q(...) notation is the Qq library: Q(t) means "an Expr that is known to have type t", and q(e) builds the expression for e while statically typechecking it. Under the hood Q(t) is literally just Expr — the type argument is a phantom annotation whose whole purpose is to let Lean typecheck your proof-building code when the tactic is compiled, rather than blowing up at runtime with a kernel error. That static typecheck is exactly what started failing after the bump.

What deriveBool returns. Its signature is deriveBool (p : Q(Prop)) : MetaM ((b : Bool) × BoolResult p b) — a dependent pair: the type of the second component depends on the value of the first. With

```lean
abbrev BoolResult (p : Q(Prop)) (b : Bool) : Type :=
  Q(Bool.rec (¬ $p) ($p) $b)
```

Bool.rec (¬p) p b is "if b then p else ¬p" written with the recursor: when b is the literal true it computes to p, when literal false to ¬p. So deriveBool q($az = $aw) returns either ⟨true, proof of az = aw⟩ or ⟨false, proof of ¬(az = aw)⟩. Crucially, when b is a variable, Bool.rec … b cannot compute — the type is "stuck" until the variable is replaced by a concrete constructor.

Why the branch needs the refined type. After let ⟨ba, ra⟩ ← deriveBool q($az = $aw), the context has ba : Bool (an opaque variable) and ra : BoolResult _ ba (stuck type). The false branch then builds q(NormNumI.ne_of_re_ne $pfz $pfw $ra), and ne_of_re_ne's last argument must be a proof of az ≠ aw — so Qq demands ra : Q(¬($az = $aw)). That only typechecks if the elaborator knows ba = false inside that branch, so it can compute Bool.rec (¬p) p false ≡ ¬p.

What "dependent match" does. A dependent match on ba substitutes the matched constructor for ba throughout the local context in each arm — so in the false arm, ra's type becomes BoolResult _ false, which reduces to Q(¬(az = aw)), and everything typechecks. A non-dependent match just branches on the value; ra keeps its stuck type BoolResult _ ba and the quotation fails.

What changed in Lean v4.32.0 — and it's the toolchain, not mathlib. I diffed mathlib v4.31.0 against v4.32.0: deriveBool and BoolResult are byte-for-byte identical. What changed is Lean's new do elaborator. From the v4.32.0 release notes: "The arms of do match are non-dependent by default — write do match (dependent := true) to recover the legacy term-match expansion." Previously, a match inside a do block was expanded into an ordinary term-level match, which is dependent; the new elaborator compiles it non-dependently for simplicity/performance, and code relying on type refinement breaks. (So the PR description's line "deriveBool now returns a dependent BoolResult" is wrong — it always did; the description should blame the do match change instead.)
 expansion." Previously, a match inside a do block was expanded into an ordinh is dependent; the new elaborator compiles it non-dependently for simplicity/performance, and code relying on type refinement breaks. (So the iveBool now returns a dependent BoolResult" is wrong — it always did; the description should blame the do match change instead.) Empirical confirmation on this exact toolchain. I ran a minimal model througject's v4.32.0: 

```lean
abbrev BR (p : Prop) (b : Bool) : Prop := Bool.rec (¬ p) p b

def f1 (p : Prop) (act : Id ((b : Bool) ×' BR p b)) : Id (PLift (p ∨ ¬ p)) := do                                                                                                       let ⟨b, h⟩ ← act
  match b with                      -- ERROR: `h` has type `BR p b`, expected `p`
  | true => return ⟨Or.inl h⟩                                                                                                                                                          | false => return ⟨Or.inr h⟩
```

- f1 (plain do match) fails with Application type mismatch: h has type BR p b but is expected to have type p — the same failure class NormNumI hit.                                  - f2, identical but with match (dependent := true) b with, compiles.
- f3, the same match written outside do-notation as a term-level match, compiles with no annotation — confirming term-mode match is still dependent by default and it's specifically the do expansion that changed.                                                                                                                                                      
Corroboration from mathlib itself: it made the identical fix in its own deriveBoolOfIff — Mathlib/Tactic/NormNum/Core.lean:154 now reads match (dependent := true) b with, where the v4.31.0 version was a plain match b with.

So for the reviewer thread: the one-line change is the minimal, officially-sis Lean release, mirroring what mathlib did internally. The only follow-up I'd suggest is editing the PR description's NormNumI bullet to attribute the breakage to the Lean 4.32 do match elaborator change rather than to deriveBool.

SpherePacking/ModularForms/EisensteinAsymptotics.lean
thefundamentaltheor3m: Seeing as it's private for now, I'm happy to leave it as is. But the lemma you pointed to is named correctly because it is namespaced. The analogue here would be something like `periodic_E₂_comp_ofComplex`. Anyway this is very minor.

thefundamentaltheor3m (approved):

PR #467, opened 2026-09-01
refactor(CauchyGoursat): assume vanishing of the top-edge integral

cameronfreer:
Weakens the decay hypothesis of the open-rectangle deformation theorems, without breaking any existing name.

## Where this sits

**Depends on:** nothing — branches from current `main`.

**Enables:** the final double-zero contour deformation, together with #466. The branches are currently independent: #466 proves `tendsto_topEdgeIntegral_zero` as a set integral over `Icc (-1) 1`; applying the new `_of_tendsto_top` variants also requires the standard conversion to an interval integral.

## The problem

Every deformation theorem in `OpenRectangular.lean` currently requires

```lean
htendsto : ∀ ε > 0, ∃ M : ℝ, ∀ z : ℂ, M ≤ z.im → ‖f z‖ < ε
```

i.e. decay uniformly over **every** real part, on an unbounded horizontal extent.

Two observations:

1. **The proofs never need it.** `htendsto` is used in exactly one place — `tendsto_integral_atTop_nhds_zero_of_tendsto_im_atTop_nhds_zero`, to show the top-edge integral vanishes — and there it is only ever instantiated at `x + y*I` for `x` in the integration interval. The decay is only ever used *on the strip* `[[x₁, x₂]]` swept by the top edge.

2. **For the intended application it is false.** The double-zero integrands look like `φ₀''(-1/z) · z² · e^{iπrz}`. Along a fixed `Im z = M`, the `z²` factor grows without bound while `|e^{iπrz}| = e^{-πrM}` is constant in `Re z`, and `φ₀''(-1/z)` degenerates as `-1/z → 0` approaches the real axis. So `‖f z‖ → ∞`, and no `M` can work. This is not a matter of proving something harder — the hypothesis cannot be discharged.

## What this adds

- **`tendsto_integral_atTop_nhds_zero_of_tendsto_im_atTop_nhds_zero_of_mem_uIcc`** — the primitive strip-local decay lemma, requiring `‖f z‖ < ε` only for `z.re ∈ [[x₁, x₂]]`. The original global-decay lemma keeps its signature as a two-line wrapper.
- **`_of_tendsto_top` variants** of the four deformation theorems, taking

  ```lean
  htop : Tendsto (fun m : ℝ ↦ ∫ x in x₁..x₂, f (x + m * I)) atTop (𝓝 0)
  ```

  directly. This is the weakest natural form: it states precisely what the proof needs, and it lets a caller supply the top-edge limit however they can prove it — strip-locally, or by any other route.

  | new | wrapper retained |
  |---|---|
  | `tendsto_integral_boundary_open_rect_one_side_atTop_of_tendsto_top` | `…_one_side_atTop_nhds_sum_other_two_sides` |
  | `integral_boundary_open_rect_eq_zero_of_tendsto_top` | `…_of_differentiable_on_off_countable` |
  | `integral_boundary_open_rect_eq_zero_of_tendsto_top'` | `…_of_differentiable_on_off_countable'` |
  | `integral_boundary_open_rect_eq_zero_of_integrable_on_of_tendsto_top` | `…_of_integrable_on` |

All four original theorems keep their exact signatures and are now one-line wrappers, so no existing call site changes.

## Not addressed here

The `sorry` at the end of the file (`…_of_integrable_on'`, deriving integrability along the second contour from the first) is a separate question and is untouched.

## Verification

Full `lake build` and the generated-import check pass, with no new `sorry` declarations. The strip-local proof uses `ε / (|x₂ - x₁| + 1)`, so zero-width and nonzero-width intervals follow the same argument. The module introduction now describes the top-edge-limit API.

### DIFF
--- a/SpherePacking/ForMathlib/CauchyGoursat/OpenRectangular.lean
+++ b/SpherePacking/ForMathlib/CauchyGoursat/OpenRectangular.lean
@@ -49,27 +49,18 @@ as `Im z → ∞` within a bounded strip while growing without bound along a hor
 lemma tendsto_integral_atTop_nhds_zero_of_tendsto_im_atTop_nhds_zero_of_mem_uIcc
     (htendsto : ∀ ε > 0, ∃ M : ℝ, ∀ z : ℂ, z.re ∈ [[x₁, x₂]] → M ≤ z.im → ‖f z‖ < ε) :
     Tendsto (fun (m : ℝ) ↦ ∫ (x : ℝ) in x₁..x₂, f (x + m * I)) atTop (𝓝 0) := by
-  wlog hne : x₁ ≠ x₂
-  · rw [ne_eq, Decidable.not_not] at hne
-    simp only [hne, integral_same, tendsto_const_nhds_iff]
+  obtain rfl | hne := eq_or_ne x₁ x₂
+  · simp only [integral_same, tendsto_const_nhds_iff]
   simp only [NormedAddGroup.tendsto_nhds_zero, eventually_atTop]
   intro ε hε
-  obtain ⟨M, hM⟩ := htendsto ((1 / 2) * (ε / |x₂ - x₁|)) <| by
-    simp only [one_div, inv_pos, Nat.ofNat_pos, mul_pos_iff_of_pos_left]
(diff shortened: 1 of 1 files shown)


cameronfreer:
awaiting-review

PR #317, opened 2026-01-28
feat(FourierExpansions): q-series infrastructure and Fourier expansion identities

cameronfreer:
## Summary

This PR completes all Fourier expansion proofs in `FourierExpansions.lean`. It provides infrastructure for proving exact Fourier expansion identities using q-series and Cauchy products.

**Relationship to #304:** PR #304 introduced the phi bounds infrastructure with sorry stubs for the Fourier expansion identities. This PR proves those identities completely. This PR is stacked on #304 and depends on it; please merge #304 first.

### Three Fourier Expansion Identities Proved

```lean
lemma E₄_sq_fourier (x : ℍ) :
    E₄ x ^ 2 = ∑' (n : ℕ), fouterm c_E₄_sq x (n + 0)

lemma E₂E₄E₆_sq_fourier (x : ℍ) :
    ((E₂ x) * (E₄ x) - (E₆ x)) ^ 2 = ∑' (n : ℕ), fouterm c_E₂E₄E₆ x (n + 4)

lemma E₄_E₂E₄E₆_fourier (x : ℍ) :
    E₄ x * (E₂ x * E₄ x - E₆ x) = ∑' (n : ℕ), fouterm c_E₄_E₂E₄E₆ x (n + 2)
```

### Polynomial Growth Bounds

- `cauchyCoeff_poly`: Cauchy product of O(n^k) × O(n^ℓ) sequences is O(n^{k+ℓ+1})
- `evenExt_poly`: Even extension preserves polynomial growth
- `toIntCoeff_poly`: toIntCoeff preserves polynomial growth
- `c_E₂E₄E₆_poly`: O(n^11) growth for (E₂E₄-E₆)² coefficients
- `c_E₄_E₂E₄E₆_poly`: O(n^10) growth for E₄(E₂E₄-E₆) coefficients
- `c_E₄_sq_poly`: O(n^9) growth for E₄² coefficients

### Cauchy Product Summability

- `norm_exp_pi_I_z_lt_one`: |exp(πiz)| < 1 for z : ℍ
- `summable_cauchy_q_series_of_poly`: generic summability for Cauchy q‑series under polynomial growth
- `cauchy_b_E₄_q_series_summable`, `cauchy_a_E₂E₄E₆_q_series_summable`, `cauchy_b_E₄_a_E₂E₄E₆_q_series_summable`: now one‑liners using the generic lemma
- `summable_evenExt_r_series`: generic summability for r‑series after even extension (used in the two Fourier proofs)
    
[...]

PR #472, opened 2026-09-01
feat(MagicFunction): polynomial bounds on Cauchy-product coefficients

cameronfreer:
Standalone coefficient definitions and polynomial-growth bounds, extracted from #317 and rebuilt directly from `main`.

## Where this sits

**Depends on:** no other PR; `CauchyCoeffBounds.lean` imports only mathlib.

**Merge order:** #472 and #477 in either order, then #304, then #473.

#304 uses the canonical `bE₄`, `bg`, and `evenCoeff` definitions and their growth bounds for its factorwise estimates. #473, stacked on #304, additionally uses the Cauchy-product machinery to prove the exact product Fourier expansions. Together, this PR and #473 replace the two parts of the now-closed #317.

## Contents

- Coefficients `bE₄` and `bg`, their convolution `cauchyCoeff`, and the even-index re-indexing `evenCoeff`.
- `exists_global_poly_bound`, shared by the convolution and re-indexing growth proofs.
- `cauchyCoeff_poly`: bounds of degrees `k` and `ℓ` give a convolution bound of degree `k + ℓ + 1`.
- Concrete bounds of degrees `11`, `10`, and `9` for the three quadratic products, with the required small-index vanishing lemmas.

The `evenCoeff` definition is unchanged: at a negative even index it returns the zeroth coefficient. The vanishing lemmas explicitly supply the hypothesis needed to make that coefficient zero.

## Verification

Full `lake build`, `lake exe mk_all --check --module`, and `git diff --check` pass. The coefficient file has no `sorry` declarations or warnings.

### DIFF
--- a/SpherePacking.lean
+++ b/SpherePacking.lean
@@ -29,6 +29,7 @@ public import SpherePacking.ForMathlib.tprod
 public import SpherePacking.MagicFunction.IntegralParametrisations
 public import SpherePacking.MagicFunction.PolyFourierCoeffBound
 public import SpherePacking.MagicFunction.a.Basic
+public import SpherePacking.MagicFunction.a.CauchyCoeffBounds
 public import SpherePacking.MagicFunction.a.Eigenfunction
 public import SpherePacking.MagicFunction.a.Integrability.ComplexIntegrands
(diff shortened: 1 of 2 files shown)


cameronfreer:
awaiting-review

PR #477, opened 2026-09-08
refactor: extract shared pointwise q-series bounds

cameronfreer:
## Stack

Based directly on `main`; a prerequisite for #304, independent of #472.
Merge this PR and #472 in either order, then #304, then #473.

## Changes

- Extract shared pointwise q-series bounds into `ForMathlib/QSeriesBounds.lean`, for shifted
  coefficients and for coefficients vanishing below `n₀`.
- Add `UpperHalfPlane.norm_le_qExpansion_of_coeff_vanish` in `ResToImagAxis.lean`. It bounds a
  1-periodic holomorphic function bounded at infinity using its canonical `qExpansion`
  coefficients, deriving summability and the series identity automatically.
- Reprove `isBigO_atImInfty_of_fourier_shift` using the shared bound, with its statement unchanged.

The coefficient-level lemmas remain available without holomorphy assumptions. The decay factor
for unshifted coefficients is `exp (-2π n₀ (im z - c))`; a short docstring remark explains why the
reference-height correction is necessary.

## Verification

Full `lake build`, `lake exe mk_all --check --module`, and `git diff --check` pass.
All three bounds use only `propext`, `Classical.choice`, and `Quot.sound`. No new sorries.

### DIFF
--- a/SpherePacking/ForMathlib/QSeriesBounds.lean
+++ b/SpherePacking/ForMathlib/QSeriesBounds.lean
@@ -5,8 +5,9 @@ Authors: Cameron Freer
 -/
 module
 
-public import Mathlib.Analysis.SpecialFunctions.Complex.Log
+public import Mathlib.Analysis.Complex.Trigonometric
 public import Mathlib.Analysis.Normed.Group.InfiniteSum
+public import Mathlib.Analysis.SpecialFunctions.Trigonometric.Basic
 import Mathlib.Tactic.Linarith
 import Mathlib.Tactic.Positivity
 
--- a/SpherePacking/ForMathlib/QSeriesBounds.lean
+++ b/SpherePacking/ForMathlib/QSeriesBounds.lean
@@ -1,0 +1,74 @@
+/-
+Copyright (c) 2026 Cameron Freer. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Cameron Freer
+-/
+module
+
+public import Mathlib.Analysis.SpecialFunctions.Complex.Log
+public import Mathlib.Analysis.Normed.Group.InfiniteSum
+import Mathlib.Tactic.Linarith
+import Mathlib.Tactic.Positivity
+
+/-!
+# Pointwise bounds for q-series
+
+Absolute summability at a reference height `c` gives uniform bounds above that height.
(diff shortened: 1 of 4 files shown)


SpherePacking/ForMathlib/Test/QSeriesBounds.lean
thefundamentaltheor3m, line 1: What is this file contributing? Unless some of these can be turned into API lemmas, this file does not seem worth merging.

SpherePacking/ForMathlib/QSeriesBounds.lean
```diff
@@ -0,0 +1,75 @@
...
+
+@[expose] public section
+
+open Real
+
+namespace Complex
+
+/-- Bound a shifted q-series above its reference height of absolute summability. -/
+lemma norm_qseries_shift_le {a : ℕ → ℂ} (n₀ : ℕ) {c : ℝ}
+    (ha : Summable fun m : ℕ ↦ ‖a m‖ * Real.exp (-(2 * π * c) * m))
```
thefundamentaltheor3m, line 35: Can this `Summable` hypothesis be stated in `fourierCoeff` language?

SpherePacking/ForMathlib/Test/QSeriesBounds.lean
cameronfreer: You're right that this wasn't useful -- though I did add a small explanation to a docstring in light of what the fifth example was getting at: [d66cd1e](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean/pull/477/commits/d66cd1eff16157bcda8e44ee5509fda7f9dc0be4) (It seems helpful to me, but I can remove that too if you prefer.)

SpherePacking/ForMathlib/QSeriesBounds.lean
cameronfreer: Yes, I checked and this works (once `a` is identified with a functions Fourier coefficients) -- see e30040c for a `qExpansion` version that derives summability from analytic assumptions; 4a53334 points to it from the docstring here. But I left the original for arbitrary coefficient sequences because it needs more assumptions for the `fourierCoeff` version. Does that seem reasonable, or should I reorganize this?

